PR #16, opened 2020-06-05
Implement option to only follow redirects on the same Host

timoles:
This pull request implements and additional flag `-follow-host-redirects` \
With this flag the tool only follows redirects as long as they're on the same host. 

---

Lets say we request "https://example.com". 
We get a redirect response to "https://example.com/test". 
The tool makes the request and we get a response with a redirect to "https://authenticate.example.com". 
Because the redirect tries to change the "Host" the tool does no longer follow the redirect.

---

This feature can, for example, be useful when scanning a range of websites which like to (eventually) redirect to an authentication portal. \
Instead of getting either (1) no redirects or (2) only redirects which point to the authentication portal we get the resources which invoke the authentication redirect. 

Implemented was the change by implementing a more complex `redirectFunc` function which compares the original host with the next to-be requested host. \
If both redirect flags are set the "follow-host-redirects" flag takes priority.

I'm happy about any feedback about the feature/implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ This will display help for the tool. Here are all the switches it supports.
 |------------------- |-------------------------------------------------------|----------------------------------------------------|
 | -H                 | Custom Header input                                   | httpx -H 'x-bug-bounty: hacker'                    |
 | -follow-redirects  | Follow URL redirects (default false)                  | httpx -follow-redirects                            |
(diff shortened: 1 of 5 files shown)


timoles:
I think this PR got closed by accident, as it was referenced in a different issue but was not merged yet. 
I changed the code according to the suggestion and also added the flag into the readme.

bauthard:
Surely this is mistake and not sure how it happened, I wil make sure we get this right, again thanks @timoles for making those changes.

Issue #17, opened 2020-06-05
Bug: -store-response not always working


timoles:
If the tool tries to write the response of a domain with a slash an error occurs. 

Steps to replicate:

```bash
# (Tested on Ubuntu 20.04) 
echo "https://www.google.com/"
httpx -l hosts.txt -json -store-response -store-response-dir ./tmp/
# This should write the file "./tmp/google.com.txt"
# > No file is created
```

Due to the trailing slash (or slashes within the path) the tool tries to write the file to the path "./tmp/google.com/.txt". `google.com` is interpreted as directory (which does not exist). This leads to an uncaught error [here](https://github.com/projectdiscovery/httpx/blob/master/cmd/httpx/httpx.go#L225-228)

PR #18, opened 2020-06-05
Fix for #17

timoles:
A fix proposal for the improper path handling when storing responses.


bauthard:
Thanks a lot, @timoles for finding and fixing this bug.

PR #20, opened 2020-06-08
Added option to include the raw response within the json output

timoles:
When the `-response-in-json` and `-json` flag are set it includes the server response within the json "serverResponse" element. 

Is the `-response-in-json` flag not set, then the "serverResponse" element is omitted and tool output is the same as before.

server_response


bauthard:
Hi @timoles, 

Thanks again for adding another option to `httpx`, Can you make sure to resolve the `conflicts`?

Mzack9999:
@timoles Thanks for your PR!

PR #30, opened 2020-07-03
Update httpx.go

bimkos:
Fix proxy


bauthard:
Thank you for the fix and PR @bimkos

PR #31, opened 2020-07-04
Update httpx.go

bimkos:
Fixed panic: runtime error: slice bounds out of range [838:33]
and add some desc to the proxy


bauthard:
Thank you for the PR @bimkos

PR #38, opened 2020-07-11
Proxy desc

bimkos:
FIX PROXY


bimkos:
Something I messed up with the commits. Tell me if there are any problems, I will delete the request. Sorry

Mzack9999:
@bimkos - Thanks for the PR! I anyway think that current naming (`HttpProxy`)  is more appropriate. Do you think it's somehow confusing?

bimkos:
I think that HttpProxy can be a bit confusing since, in fact, you can use any proxy (socks5, HTTP, etc.).

bauthard:
Hi @bimkos, 

Thank you for the feedback, but as this is something already in use and published already, we should keep it as is it, as such closing this PR, thank you again.

PR #50, opened 2020-07-23
request path option

Mzack9999:


bauthard:
```bash
> chaos -silent -d hackerone.com  | ./httpx -silent -status-code -path /api -status-code

https://mta-sts.hackerone.com [404]
https://mta-sts.managed.hackerone.com [404]
https://mta-sts.forwarding.hackerone.com [404]
https://docs.hackerone.com [404]
https://support.hackerone.com [302]
https://api.hackerone.com [404]
https://resources.hackerone.com [301]
https://www.hackerone.com [403]
```

I can confirm it's working as expected, just need to update the requested URL in output as well.

bauthard:
@Mzack9999 the path also needs to update for the **txt file** created with `store-response` flag, as of now file gets created with domain name even the response saved is for the path.

PR #56, opened 2020-08-01
fixed setting custom headers.

n3k:
Previous to this, passing a header like User-Agent as the following: "User-Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:47.0) Gecko/20100101 Firefox/47.0", trimed the value to "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,6 @@ RUN go get -u github.com/golang/dep/cmd/dep
 WORKDIR /go/src/app
 
 # Install
-RUN go get -u github.com/projectdiscovery/httpx/cmd/httpx
+RUN go get -u github.com/n3k/httpx/cmd/httpx
 
 ENTRYPOINT ["httpx"]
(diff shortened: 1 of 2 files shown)


bauthard:
Thank you for the PR @n3k, 

Can you clarify what issue this PR trying to solve? we are not doing anything for `User-agent` which is user-controlled.

bauthard:
Closing this, as the purpose of this PR is not clear.

n3k:
You're splitting the headers by ":" , some User-Agents have ":" as part of it, and so my fix takes that into account. 
Again, if you try to set a user agent like this one: "User-Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:47.0) Gecko/20100101 Firefox/47.0" .. Given the string contains a ":", it will only keep the string up to "User-Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv". Try it out yourself.

n3k:
> Closing this, as the purpose of this PR is not clear.

So is this going to be included or not?.. don't you see a single line fixes the problem of strings containing ":" ?

Haones:
@bauthard I confirmed as well that if you pass a header with multiple colons it trims out anything after second colon.

PR #75, opened 2020-08-25
all: simplify logic statements and remove unneeded conversions

jimen0:
Hi Project Discovery team,

This PR does the following:

- Removes unneeded negative length checks for slices
- Removes unneeded string to string conversions
- Fixes a typo 🤷 
- Simplifies if/else statements in methods that return a bool

All commits are signed and I tested my changes using both Go 1.15 and `tip` branch. Please, let me know if any change is needed from my side.

Best,
Miguel

### DIFF
--- a/common/cache/dialer.go
+++ b/common/cache/dialer.go
@@ -39,7 +39,7 @@ func NewDialer(options Options) (DialerFunc, error) {
 				finalIps = append(finalIps, ip)
 			}
 		}
-		if err != nil || len(finalIps) <= 0 {
+		if err != nil || len(finalIps) < 0 {
 			return nil, &NoAddressFoundError{}
 		} // Dial to the IPs finally.
 		for _, ip := range ips {
--- a/common/cache/dialer.go
+++ b/common/cache/dialer.go
@@ -39,7 +39,7 @@ func NewDialer(options Options) (DialerFunc, error) {
 				finalIps = append(finalIps, ip)
 			}
 		}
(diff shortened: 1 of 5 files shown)


Mzack9999 (approved):

common/cache/dialer.go
```diff
@@ -39,7 +39,7 @@ func NewDialer(options Options) (DialerFunc, error) {
 				finalIps = append(finalIps, ip)
 			}
 		}
-		if err != nil || len(finalIps) <= 0 {
+		if err != nil || len(finalIps) < 0 {
```
Ice3man543, line 42: Could you change finalIps to `<=0`since it can never be less than 0. Here we are checking if there are no IP addresses returned from the resolver.

Ice3man543 (changes requested):

common/cache/dialer.go
jimen0: Hi @Ice3man543,

I updated the PR to check if length of finalIps is exactly 0. It doesn't make sense to check if it's less than 0. When building my PR I mistyped less than 0 instead of equals to 0. Sorry about that.

Let me know if any other change is needed from my side.

Ice3man543:
@jimen0 thank you for the changes. At first i thought it was an issue from our side, :P. Typos happen. Merging the PR now.

Ice3man543 (approved):
Thanks for the PR! Merging.

PR #86, opened 2020-08-31
fix and improve method CLI option

Mzack9999:


bauthard (approved):

bauthard:
This is tested and verified, now `-x all` probes all the supported method, and this also fixes the default to print the request method.

PR #108, opened 2020-09-14
adding cname + ips lookup in global cache

Mzack9999:


bauthard:
`cname` flag is missing in options.

bauthard (dismissed):
adding `cname` flag in options.

bauthard:
This is now working, I also wanted to check if we want to print all the **cnames**? like we doing now, unlike **IPs** where we print 1st IP, maybe we can do the same with **cname** as well? 

```
https://www.hackerone.com
https://api.hackerone.com
https://mta-sts.hackerone.com [hacker0x01.github.io]
https://mta-sts.managed.hackerone.com [hacker0x01.github.io]
https://support.hackerone.com [hackerone.zendesk.com]
https://mta-sts.forwarding.hackerone.com [hacker0x01.github.io]
https://docs.hackerone.com [hacker0x01.github.io]
https://resources.hackerone.com [read.uberflip.com,nlb-ext-traefik-ca-central-1-d39d611502919b07.elb.ca-central-1.amazonaws.com]
```

PR #110, opened 2020-09-14
adding string/regex match/filter

Mzack9999:


bauthard:
@Mzack9999 Can you share an example for this? Following command return blank. 

```
$ echo hackerone.com | ./httpx -match-string hackerone -silent
```

bauthard:
This needs more testing, as this is still not working in the expected way.

Issue #117, opened 2020-09-25
Add golangci-lint to the current GitHub workflow


vzamanillo:
`golangci-lint-action` It's the official `GitHub` action for `golangci-lint` from it's authors. The action runs `golangci-lint` and reports issues from linters.
`
golangci-lint` is a fast Go linters runner. It runs linters in parallel, uses caching, supports yaml config, has integrations with all major IDE and has dozens of linters included.

Including this action into the workflow would increment the project code quality and could prevent possible future leaks and/or failures setting a minimum of checks and rules, I think it could be very favorable.

bauthard:
Added here https://github.com/projectdiscovery/httpx/pull/121

PR #124, opened 2020-09-29
Fixed reading input file with -l

### LINKED ISSUE #123: Issue with file read input flag
Reproduce with:

```sh
$ echo -e "example.com\ngoogle.com" > x
$ httpx -l x
```

Expected:

```
http://example.com
https://exampe.com
http://google.com
https://google.com
```

Outcome: none

Obtain expected outcome with:

```sh
$ httpx < x
```

vzamanillo:
Fixes #123 but is not related to #122.

This bug was introduced on #117, closing the input file has to be deferred or we will try to read from a closed stream.

### DIFF
--- a/common/fileutil/fileutil.go
+++ b/common/fileutil/fileutil.go
@@ -24,8 +24,10 @@ func HasStdin() bool {
 		return false
 	}
 
-	isPipedFromChrDev := (stat.Mode() & os.ModeCharDevice) == 0
-	isPipedFromFIFO := (stat.Mode() & os.ModeNamedPipe) != 0
+	mode := stat.Mode()
+
+	isPipedFromChrDev := (mode & os.ModeCharDevice) == 0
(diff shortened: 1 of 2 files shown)


common/fileutil/fileutil.go
```diff
@@ -24,8 +24,10 @@ func HasStdin() bool {
 		return false
 	}
 
-	isPipedFromChrDev := (stat.Mode() & os.ModeCharDevice) == 0
-	isPipedFromFIFO := (stat.Mode() & os.ModeNamedPipe) != 0
+	mode := stat.Mode()
+
+	isPipedFromChrDev := (mode & os.ModeCharDevice) == 0
+	isPipedFromFIFO := (mode & os.ModeNamedPipe) != 0
```
vzamanillo, line 30: This changes can be replicated to other projects (subfinder, naabu and nuclei) to fix an small delay while checking stdin.

PR #133, opened 2020-10-07
Dockerfile updated, image size reduced using multi-stage build

jsav0:
- Reduced docker image from 525MB to 16.9MB using multi-stage build.


bauthard:
Thank you @jsav0 for the update.

PR #157, opened 2020-11-08
resolves teichopsia/httpx#1 - enhanced tls tracking

teichopsia:
no idea if this is of any value to the mainline. added it to solve specific use cases I had and it served as a forced crash course in golang. flames welcome. love the toolchain otherwise.

the problem I was trying to solve was basically:

```
cat inventory-cert-list | httpx -srtls -srd snuffy -sr -tls-probe -fulljson | jq ....
```

and end up with targetN.txt and targetN.tls where targetN.tls contained the tls parameters encountered and the PEM chain involved. -srtls solves that part.

the other issue is how to deal with the -json equivalent output. because adding the cert subtree is 'verbose' and there were use cases that wanted it present/omitted alike i added '-fulljson' to invoke -json behavior yet allow folding the cert subtree. yes you could filter this with a convoluted jq remap but due to the depth it was easier to handle in httpx before jq.

I probably chose an incomplete/poor solution by patching json marshalling but like I said, first stab at golang. the initial documentation for json tagging fields only omits if the RHS is not present. if your LHS is an array/slice it can't undo the key added so you have to filter before encountering it hence juggling TLSData around so much.

the entire use case for this is chasing down stray vips with stale/mismatched certificates. when I capture the host, sni, ip I can cross-reference against dns lints to fully enumerate pachinko mismatches in an environment. nothing worse than getting a weird response and wishing you had the tls environment for that request when it happened instead of chasing it down manually later

### DIFF
--- a/common/httpx/tls.go
+++ b/common/httpx/tls.go
@@ -1,17 +1,59 @@
 package httpx
 
 import (
+	"crypto/tls"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"github.com/projectdiscovery/httpx/common/cache"
 	"net/http"
 )
 
+type CertificateBlock struct {
+	Envelope map[string]string `json:"env,omitempty"`
+	Payload  []byte            `json:"payload,omitempty"`
+}
+
+type CertificateFold struct {
+	Certificate []CertificateBlock `json:"cert,omitempty"`
+	CertFold    bool               `json:"-"`
+}
+
 // TLSData contains the relevant Transport Layer Security information
-type TLSData struct {
+type TLSDataCore struct {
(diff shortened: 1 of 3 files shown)


bauthard:
Hi @teichopsia, 

Apologies for the delay, and thanks a lot for the PR, we reviewed and discussed these changes, as you know our primary aim to keep the tool simple and focused on extracting the base data, for post-processing we can work on new tools if needed that works along with all of our tools without adding those capabilities to the same tool with more options and we already have a lot of them, so by keeping base features and simplicity at priority, we suggest you keep working on this on the forked version of yours as we can't merge this into the master, with that closing this PR, thank you for sharing for your ideas in this PR.

PR #175, opened 2020-12-14
support extract info by custom regex

b3r1ch:
hi @team-projectdiscovery 
i add a new flag `-costom-regex` to extract information(string) by custom regex
this flag default value is empty,if you wanna use custom regex, just set `-costom-regex` like this:

```bash
cat domains.txt | ./httpx -path /siteInfo -custom-regex "(?im)\"title\"\:\"(.*?)\""
```

then httpx will print match string by regex  `(?im)"title":"(.*?)"`

I simply realized the idea of supporting custom regex,i hope this pull request will help httpx to be better

### DIFF
--- a/common/httpx/customregex.go
+++ b/common/httpx/customregex.go
@@ -1,0 +1,16 @@
+package httpx
+
+import (
+	"regexp"
+	"strings"
+)
+
+// ExtractInfo by custom regex from a response
+func ExtractInfoByCustomRegex(r *Response, regex string) (info string) {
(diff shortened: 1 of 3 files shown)


team-projectdiscovery:
Thank you for taking your time to submit a PR for this, as listed in the issue section, this is something that fits better with the nuclei in terms of use case, on another side you can still run regex on the JSON response data with `jq`

Reference: https://stackoverflow.com/questions/37328327/apply-regex-on-jq, with that said closing this PR.

PR #181, opened 2020-12-20
Improved title extraction

Mzack9999:


bauthard:
This looks intended as per response data, new lines around the title.

 ```bash
 echo 124.127.92.139:8088 | ./httpx -title -silent
http://124.127.92.139:8088 [
	南京万川医药-国家电子监管码系统（Varsal SDA）
]
```

PR #190, opened 2021-02-10
fixed build error

jsav0:
`go get -u` was causing the module dependencies to be updated (`-u`) ignoring the version declarations in your project's `go.mod` file thus causing a build error.


ehsandeep (approved):

ehsandeep:
Thanks again @jsav0

Issue #63, opened 2020-08-08
[BUG] CIDR in combination with -ports


marcopolo157:
## To replicate the issue of missing valid host with CIDR input

with `prips`:- 
```bash
> prips 1.1.1.0/24 | httpx -title -content-length -status-code -ports 80,443 -silent | grep 1.1.1.1:80

http://1.1.1.1:80 [301] [186] [301 Moved Permanently]
```

with `httpx` internal CIDR handler: -

```bash
> echo 1.1.1.0/24 | httpx -title -content-length -status-code -ports 80,443 -silent | grep 1.1.1.1:80
```

## To replicate the duplication issue with CIDR input

```bash
> echo 1.1.1.0/24 | httpx -title -content-length -status-code -ports 80,443 -silent | sort | grep http://1.1.1.24:80
http://1.1.1.24:80 [403] [16] []
```

```bash
echo 1.1.1.0/24 | httpx -title -content-length -status-code -ports 80,443 -silent | sort | grep http://1.1.1.24:80

http://1.1.1.24:80 [403] [16] []
http://1.1.1.24:80 [403] [16] []
http://1.1.1.24:80 [403] [16] []
```
------------

I have opened ports 80 and 443 on 192.168.8.1
```
echo 192.168.8.1 | ./httpx  -title -content-length -status-code -ports 80,443

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   / 
 / / / / /_/ /_/ /_/ /   |  
/_/ /_/\__/\__/ .___/_/|_|  
             /_/              v1           

		projectdiscovery.io

[WRN] Use with caution. You are responsible for your actions
[WRN] Developers assume no liability and are not responsible for any misuse or damage.
https://192.168.8.1:443 [307] [13] []
http://192.168.8.1:80 [307] [13] []
```

But when I run CIDR scan, the httpx does not return any open port on 192.168.8.1
```
echo 192.168.8.0/24 | ./httpx  -title -content-length -status-code -ports 80,443

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   / 
 / / / / /_/ /_/ /_/ /   |  
/_/ /_/\__/\__/ .___/_/|_|  
             /_/              v1           

		projectdiscovery.io

[WRN] Use with caution. You are responsible for your actions
[WRN] Developers assume no liability and are not responsible for any misuse or damage.
```

[...]

PR #203, opened 2021-02-19
fix duplicated result when use CIDR targets and -ports

mashaz:
## Describe

When use CIDR targets and -ports, result maybe duplicated cause the parameter ```target``` pass to ```analyze()``` is duplicated.

## Replicate

Add ```fmt.Println(domain)``` in analyze function and execute ```echo CIDR | ./httpx  -status-code -threads 50 -title  -ports 80```, it's very clear and easy to replicate.

I think it's related to #63 .


ehsandeep:
thank you @mashaz for the fix.

PR #212, opened 2021-03-10
Update options.go

geeknik:
typo fix s/lenght/length/g


ehsandeep:
Thank you @geeknik, we fixed it in the development version.

PR #247, opened 2021-04-27
 Fix ip/cname resolve bug

iflody:
When the url's format like: https://www.google.com:443, the domain pass
 to the resolver will be www.google.com:443 and can't get the right
 result

 This patch split the port at the start of the function and rewrite the
 domain to the right format, which makes the result of IP/Cname be
 right.

Example:

echo https://www.baidu.com:443 | httpx -ip -cname

The result will be:

https://www.baidu.com:443 []

After fix it will be:

https://www.baidu.com:443 [104.193.88.77] [www.a.shifen.com]


ehsandeep:
Works as described in the PR:-

```
httpx $echo http://www.baidu.com:443 | httpx -ip -cname -silent
http://www.baidu.com:443 []

httpx $echo http://www.baidu.com:443 | ./httpx -ip -cname -silent
http://www.baidu.com:443 [104.193.88.123] [www.a.shifen.com]
```

Mzack9999 (approved):

ehsandeep:
Thank you for the fix and PR @iflody

PR #251, opened 2021-04-27
Adding support for redirect chain dump

Mzack9999:


ehsandeep:
@Mzack9999 New additions working as expected, with unexpected change.

 with change, `-sr` flag does not save the response body in the output file, to replicate this behavior, run below cmd from old and binary of this branch and inspect the output file.
 
 ```
 echo hackerone.com | httpx -follow-redirects -silent -sr
 ```
 
  ```
 echo hackerone.com | ./httpx -follow-redirects -silent -sr
 ```

ehsandeep:
@Mzack9999 highlighted issue is now fixed, I also noticed the default JSON output contains `chain` data in the JSON response, can we sync with `store-chain` flag to keep the default JSON output minimal.

```
./httpx -follow-redirects -silent -json | jq .
```

PR #307, opened 2021-06-16
[Edge Case] Fixing gzip header server misconfiguration

Mzack9999:
Some servers specify the `Content-Encoding: gzip` header within the response, but the body is not compressed. Although it's s server misconfiguration, we try to perform the request again.


Ice3man543 (approved):
lgtm

PR #321, opened 2021-07-15
json output

LuitelSamikshya:


Mzack9999:
**New Flag `status`**
Add a new flag `status` that will print successul and failed probes to stdout as follows:
```
$ cat a.txt
hackerone.com
aaaahackerone.com
$ cat a.txt | httpx -status
https://hackerone.com [success]
http://aaaahackerone.com [failure]
```

**JSON output with `status` flag**
If both `status` and `json` are specified the output should change from:

```
$ echo aaaahackerone.com | go run . -verbose -debug
[DBG] Failure 'http://aaaahackerone.com': GET http://aaaahackerone.com giving up after 1 attempts: Get "http://aaaahackerone.com": could not resolve host
```

To:

```
$ echo aaaahackerone.com | go run . -status -json
{url:"http://aaaahackerone.com","original-input":"aaaahackerone.com","failed": true, "error":"could not resolve host"} 
```

runner/runner.go
```diff
@@ -615,7 +615,7 @@ retry:
 			retried = true
 			goto retry
 		}
-		return Result{URL: URL.String(), err: err}
+		return Result{OriginalInput:  domain, Timestamp: time.Now(), Failed: true}
```
Mzack9999, line 618: should be:
```
return Result{URL: URL.String(), OriginalInput: domain, Timestamp: time.Now(), Failed: true, err: err}
```

runner/runner.go
```diff
@@ -932,6 +933,7 @@ retry:
 
 // Result of a scan
 type Result struct {
+	OriginalInput	 string		`json:"original-input"`
```
Mzack9999, line 936:
```
OriginalInput string `json:"original-input,omitempty"`
```

runner/runner.go
```diff
@@ -967,6 +969,7 @@ type Result struct {
 	Technologies     []string          `json:"technologies,omitempty"`
 	Chain            []httpx.ChainItem `json:"chain,omitempty"`
 	FinalURL         string            `json:"final-url,omitempty"`
+	Failed			 bool			   `json:"failed"`
```
Mzack9999, line 972:
```
Failed bool `json:"failed,omitempty"`

File: runner/runner.go [hunk 4]
```diff
@@ -367,7 +367,7 @@ func (r *Runner) RunEnumeration() {
 			if resp.err != nil {
 				gologger.Debug().Msgf("Failure '%s': %s\n", resp.URL, resp.err)
 			}
-			if resp.str == "" {
+			if resp.str == "" && !resp.Failed{
```
Mzack9999, line 370: As per the last update also failed responses should be printed out, so here it should be:
```
if resp.str == "" {
```

Mzack9999:
Superseded by #323

PR #330, opened 2021-07-26
Issue#315-rate-limit-support

LuitelSamikshya:


runner/runner.go
```diff
@@ -426,10 +436,12 @@ func (r *Runner) RunEnumeration() {
 			for _, p := range r.options.requestURIs {
 				scanopts := r.scanopts.Clone()
 				scanopts.RequestURI = p
+				r.ratelimiter.Take()
```
Mzack9999, line 439: I think we can remove it here as it's already called in the `r.process` function (line 467)

runner/runner.go
```diff
@@ -426,10 +436,12 @@ func (r *Runner) RunEnumeration() {
...
 				r.process(t, &wg, r.hp, protocol, scanopts, output)
 				reqs++
 			}
 		} else {
+			r.ratelimiter.Take()
```
Mzack9999, line 444: I think we can remove it here as it's already called in the `r.process` function (line 467)

Mzack9999 (changes requested):

runner/runner.go
LuitelSamikshya: I ran some  tests removing this line but RPS seems to exceed the rate limit.

runner/runner.go
LuitelSamikshya: I ran some  tests removing this line but RPS seems to exceed the rate limit.

Mzack9999 (approved):

PR #336, opened 2021-08-03
Sort technologies before output to make diffing results easier

jonare:
Technologies presented in a predictable order makes diffing to previous scans easier. It can also be easier to spot outliers within a single list of results when the same tech stack is displayed consistently.


Mzack9999 (approved):

ehsandeep:
Thank you @jonare for this update.

PR #338, opened 2021-08-04
[Feature] Adding support to scan only ports 80/443 if the ip is part of CDN range

Mzack9999:


ehsandeep (approved):

ehsandeep:
Ready to merge after resolving git conflict.

PR #340, opened 2021-08-05
Adding support for non-ascii characters in path for unsafe requests

Mzack9999:


ehsandeep (approved):

ehsandeep:
Now non-ascii characters are supported with `unsafe` flag.

```
echo hackerone.com | ./httpx -path "/%hell" -unsafe -silent
https://hackerone.com/%hell
```

Issue #337, opened 2021-08-03
can not grab tls


leveryd:
**Describe the bug**

can not grab tls on "45.120.31.26"
```
echo 45.120.31.26 | httpx -json  -silent --timeout=20 --tls-grab
```

`curl https://45.120.31.26 -k` and 443 port is open

**Environment details**

- `httpx -version`: v1.1.0
- `go version`: go1.15.5 linux/amd64
- `uname -a`: root

**Error details**
![image](https://user-images.githubusercontent.com/1846319/128006717-3348f0b0-758c-4eb2-97ee-3c7c32e91b37.png)

ehsandeep:
```
curl https://45.120.31.26 -k
curl: (92) HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
```

@leveryd I get the following error, but can confirm this URL are is accessible on 443 that httpx failed to connect due to an unhandled error. 

```
echo 45.120.31.26 | httpx -debug

[DBG] Failure 'http://45.120.31.26': GET http://45.120.31.26 giving up after 1 attempts: Get "http://45.120.31.26": EOF
```

Mzack9999:
`httpx` actually only returns results for valid HTTP servers or error otherwise. Not sure if `tls-grab` shall be coupled with `-probe` always to collect results.

ehsandeep:
Hello, @leveryd.

While we wanted to fix/support this edge case, we discovered that doing so would break a few intended things for httpx, so we decided not to fix it in httpx and instead will create a new/dedicated project for SSL grabbing called **tlsx**. Please contact us at https://discord.gg/projectdiscovery to suggest a feature or use case for SSL scanning that you wanted to see implemented.

We're closing this issue since we'll be dealing with it in another project.

PR #341, opened 2021-08-05
Adding support for tls data extraction on connection error

Mzack9999:


ehsandeep:
As per - https://github.com/projectdiscovery/httpx/issues/337#issuecomment-894021335

PR #345, opened 2021-08-06
Adding stop-resume support

Mzack9999:


ehsandeep (approved):
The feature is working as expected.

Minor update:- we can include information similar to dnsx upon scan cancellation and resume.

PR #365, opened 2021-08-16
Adding support for maximum redirects number

Mzack9999:


ehsandeep (approved):

Mzack9999:
Wont Fix: the base `net/http` library returns nil response for infinite redirect loops that hits a critical timeout error. For now the only solution is to use a large `timeout` value. This limitation has been removed in the rawhttp version at https://github.com/projectdiscovery/rawhttp/pull/24/commits/1c76ed1410f068cb040b7a899093f88c3809391c

PR #369, opened 2021-08-16
Improving case behavior with methods CLI option

Mzack9999:


ehsandeep (commented):
```
httpx $echo hackerone.com | ./httpx -x head -content-length -silent
https://hackerone.com [0]
httpx $echo hackerone.com | ./httpx -x HEAD -content-length -silent
https://hackerone.com [0]
httpx $echo hackerone.com | ./httpx -x head -content-length -silent -unsafe
https://hackerone.com [155]
httpx $echo hackerone.com | ./httpx -x get -content-length -silent
https://hackerone.com [92]
httpx $echo hackerone.com | ./httpx -x GET -content-length -silent
https://hackerone.com [92]
```

looks like HEAD method is not supported after change without unsafe.

ehsandeep (approved):

PR #403, opened 2021-09-16
Fixing POST requests with body

LuitelSamikshya:


runner/runner.go
```diff
@@ -726,8 +725,12 @@ retry:
 		}
 	} else {
 		// Create a copy on the fly of the request body
-		bodyBytes, _ := req.BodyBytes()
-		req.Request.Body = ioutil.NopCloser(bytes.NewReader(bodyBytes))
+		var bodyBytes []byte
```
Mzack9999, line 728: I think it's possible to remove `bodyBytes` and use directly `scanopts.RequestBody`

Mzack9999 (changes requested):

Mzack9999 (approved):

Mzack9999:
- [ ] Add regression test for **POST request with body**

Mzack9999 (approved):

PR #423, opened 2021-10-18
Adjust url output when customizing the host request header

z0neGit:
Before:

`echo 1.2.3.4:8008 | httpx -H "Host: www.test.com"`
----> url https://1.2.3.4:8008

After:

`echo 1.2.3.4:8008 | httpx -H "Host: www.test.com"`
----> url https://www.test.com:8008

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -840,6 +840,12 @@ retry:
 	} else if !stringsutil.HasSuffixAny(fullURL, scanopts.RequestURI) {
 		parsedURL.RequestURI = scanopts.RequestURI
 	}
+	if r.options.CustomHeaders.Has("Host:") {
+		hpCustomHost := strings.Split(hp.CustomHeaders["Host"], ":")[0]
+		if parsedURL.Host != hpCustomHost {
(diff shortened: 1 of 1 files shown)


Mzack9999:
@z0neGit Thanks for this PR. Could you elaborate on the use case of this change? The printed URL should represent the host that was tested, that, as per your example, has IP address `1.2.3.4:8008` and not `www.test.com`. If reused later, would resolve to a different host than the original one.

z0neGit:
> @z0neGit Thanks for this PR. Could you elaborate on the use case of this change? The printed URL should represent the host that was tested, that, as per your example, has IP address `1.2.3.4:8008` and not `www.test.com`. If reused later, would resolve to a different host than the original one.

@

z0neGit:
@Mzack9999 TThank you for your review. This example is used for host collision. You may not get any response when sending packets directly to `1.2.3.4:8008`, but you will get a response when adding the host header flag of `www.test.com`, this site must be bound with the host header before it can be accessed. Like this [project](https://github.com/fofapro/Hosts_scan).

PR #452, opened 2021-12-06
Remove * and/or . (prefix) from target

### LINKED ISSUE #447: TLS probe discards entries starting with *
### httpx version:
dev

### Current Behavior:
httpx rejects / discard input if it contains or starts with `*` / `.` in it - `*.answers.search.yahoo.com`

### Expected Behavior:
probing - answers.search.yahoo.com

### Steps To Reproduce:

```console
echo in.search.yahoo.com | httpx -silent -json -tls-grab | jq -r .'"tls-grab".dns_names[]' | sed 's/^*.//' | grep -e "\." | httpx -silent | wc -l
61

echo in.search.yahoo.com | httpx -silent -tls-probe | wc -l
45
```

reported by detonXX

parrasajad:
to fix #447

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -653,15 +653,24 @@ func (r *Runner) process(t string, wg *sizedwaitgroup.SizedWaitGroup, hp *httpx.
 						if scanopts.TLSProbe && result.TLSData != nil {
 							scanopts.TLSProbe = false
 							for _, tt := range result.TLSData.DNSNames {
+								if !r.testAndSet(tt) {
+									continue
+								}
 								r.process(tt, wg, hp, protocol, scanopts, output)
 							}
 							for _, tt := range result.TLSData.CommonName {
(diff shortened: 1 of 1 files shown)


ehsandeep (changes requested):
duplicate input check in missing like httpx does in other cases.

```
$echo in.search.yahoo.com | ./httpx -silent -tls-probe | wc -l
78

$echo in.search.yahoo.com | ./httpx -silent -tls-probe | sort -u | wc -l
62
```

Mzack9999:
We can reuse the helper function added to stream modality to avoid dupes: https://github.com/projectdiscovery/httpx/blob/d7e081c08a1e16a79c20f40a1b20e21f2c10d151/runner/runner.go#L320 can be used

Mzack9999 (approved):

ehsandeep (approved):

PR #471, opened 2022-01-03
Adding CLI option to trim default HTTP/HTTPS ports from Host header

Mzack9999:
This PR adds the `-leave-default-ports` CLI option. If specified, avoid trimming `:80` for HTTP and `:443` for HTTPS from the Host header


LuitelSamikshya (approved):
Looks good.

ehsandeep (changes requested):
This feature seems to be supported already for port `80,443` with the default run but not when using `ports` flag, we can extend the default support with `ports` flag to cover this case.

ehsandeep (approved):

PR #476, opened 2022-01-03
Add support for favicon hash

Mzack9999:
This PR supports favicon extraction with a new flag `-favicon`. If used, the path will be modified to `/favicon.ico`. The resulting hash can be matched or filtered with:
```
-mfc, -match-favicon    Match Favicon hash
-ffc, -filter-favicon    Filter Favicon hash
```


ehsandeep (dismissed):
We can mimic the hash calculation similar to [shodan favicon](https://faviconmap.shodan.io) to make it reusable and map new hashes.

ehsandeep (approved):

Ice3man543 (approved):

Issue #496, opened 2022-01-13
-cdn -json should return true or false


brenocss:
### httpx version:
v1.1.5

### Current Behavior:

```bash
echo 'google.com' | httpx -cdn -json | jq .cdn
```
`null`

### Expected Behavior:

```bash
echo 'google.com' | httpx -cdn -json | jq .cdn
```
`false`

### Use case
```bash
echo 'google.com
hackerone.com' | httpx -json >> log.json

echo 'google.com
hackerone.com' | httpx -cdn -json >> log.json
```
```bash
# Get all hosts's that is from any cdn
cat log.json | jq 'select(.cdn==true).url'
```

```bash
# Get all hosts's that isn't from any cdn
cat log.json | jq 'select(.cdn==false).url'
# cat log.json | jq 'select(.cdn!=true).url' doesn't work because it returns null with or without the -cdn flag, getting invalid results 
```

ehsandeep:
@brenocss this feature will be updated with https://github.com/projectdiscovery/httpx/issues/492, so instead of returning true/false, it will return the name of the cdn.

PR #499, opened 2022-01-15
Expose CDN Name in output and json

Ice3man543:


ehsandeep (changes requested):
```console
echo https://www.hackerone.com | ./httpx -silent -cdn

https://www.hackerone.com [projectdiscovery]
```

brenocss:
The cdn field shouldn't be omitted when false. https://github.com/projectdiscovery/httpx/issues/496
```json
"cdn": false
```

ehsandeep (approved):

Mzack9999 (approved):

PR #500, opened 2022-01-15
Added stats-interval for changing stats display frequency

Ice3man543:


ehsandeep (changes requested):
We can enable the `stats` flag internally if `si` flag is used without stats flag.

ehsandeep (approved):

PR #502, opened 2022-01-17
Get a notification once the process is done

AbhijithAJ:
I routinely verify if this command is complete when I execute it.
As a result, I've created an [intimations](https://github.com/AbhijithAJ/intimations) module that will send me a telegram message once it's completed.

Hope this will help others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ https://support.hackerone.com
 ```
 
 ##### TIP 
-Using the python intimations module, you can receive a message to your telegram mobile application once the execution is complete.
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @AbhijithAJ, thanks for letting me know about this project, additonally we have also published [notify](https://github.com/projectdiscovery/notify) including support for multiple providers, with that said it's not something we wanted to include in each proejct readme.

AbhijithAJ:
Thanks for the response.

PR #521, opened 2022-02-18
detect favicon

LuitelSamikshya:
This PR automatically adds favicon-mmh3 to json output if received response's content-type has "image".

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1202,7 +1202,8 @@ retry:
 	}
 
 	var faviconMMH3 string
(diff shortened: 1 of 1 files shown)


Mzack9999 (changes requested):
After checking the current implementation, I believe that the JSON output should always contain all hashes, regardless of the `content-type` or the flag options, used only for the classical CLI output. What do you think, @ehsandeep?

ehsandeep:
Closing in favor of https://github.com/projectdiscovery/httpx/issues/525

PR #528, opened 2022-02-25
hash improvement

LuitelSamikshya:


ehsandeep (changes requested):
There is slight change on mmh3 hash genration with `favicon` flag vs the default one, the default one can be adoped to mimic the pythoic (`-favicon`) mmh3 hash.

```console
echo https://mta-sts.forwarding.hackerone.com/favicon.ico | ./httpx -hash mmh3 -silent
https://mta-sts.forwarding.hackerone.com/favicon.ico [2594644036]

echo https://mta-sts.forwarding.hackerone.com | ./httpx -favicon -silent
https://mta-sts.forwarding.hackerone.com/favicon.ico [-1700323260]
```

ehsandeep (changes requested):
As it's unclear the types of hashes are supported, this information can be made available to help menu by printing `hashType`

```
-hash string         Display response body hash (supported: md5,mmh3,sha256,sim)
```

M09Ic:
> ```
> https://mta-sts.forwarding.hackerone.com/favicon.ico
> ```

This bug is because I forgot to convert the final mmh3 hash to int32.
 
just change:
`common/hashes/Mmh3` 
```
func Mmh3(data []byte) string {
	var h32 = murmur3.New32WithSeed(0)
	h32.Write(stdBase64(data))
	return fmt.Sprintf("%d", int32(h32.Sum32()))
}
```

test:
`echo "https://www.bilibili.com/" | httpx -favicon -hash mmh3`

`https://www.bilibili.com/favicon.ico [-1825654886] [-1825654886]`

ehsandeep (approved):

Mzack9999 (approved):

PR #585, opened 2022-04-18
fix #584

### LINKED ISSUE #584: datarace in `Runner.process`: range value `target` is captured (by-reference) in goroutine
In the method `process` of `Runner`, the range value `target`:
https://github.com/projectdiscovery/httpx/blob/9113a3be17f01253cc7ae7f30e161a0024117591/runner/runner.go#L659-L665
is captured by in a goroutine
https://github.com/projectdiscovery/httpx/blob/9113a3be17f01253cc7ae7f30e161a0024117591/runner/runner.go#L712-L714

This leads to a data race between the `for-range` and gouroutines accessing `target`

(Found with [`revive`](https://github.com/mgechev/revive))

chavacava:
fix #584 by passing range value `target` as parameter of the goroutine


Mzack9999 (approved):

Mzack9999:
@chavacava Thanks for opening this PR and nice catch!

PR #588, opened 2022-04-18
Save 8 bytes per runner.Result allocation

jimen0:
TL;DR: `runner.Result` wasn't aligned in an optimal way leading to 8 bytes being wasted per object. Since it stores results, I believe this is an easy win that makes `httpx` a little bit less resource-intensive when working on large jobs.

```console
$ fieldalignment github.com/projectdiscovery/httpx/runner       
[REDACTED]/httpx/runner/runner.go:1437:13: struct of size 600 could be 592
```

Below you'll find the struct layout before and after. Padding (wasted, unused bytes) is seen in red color.

* Old struct

    result-unoptimized

* Optimized struct

    result-optimized

[`fieldalignment`](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment) is an official Go tool.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -597,7 +597,7 @@ func (r *Runner) RunEnumeration() {
 			if len(r.options.OutputFilterCdn) > 0 && stringsutil.EqualFoldAny(resp.CDNName, r.options.OutputFilterCdn...) {
 				continue
 			}
-			
+
 			row := resp.str
 			if r.options.JSONOutput {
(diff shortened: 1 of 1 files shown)


jimen0:
FYI @ehsandeep - Please, let me know your thoughts on this! This is just the struct layout change + `gofmt -w -s`.

Mzack9999 (approved):

PR #625, opened 2022-05-16
cname with ports flag bug-fix

LuitelSamikshya:
This PR fixes a bug where cname is not displayed with ports flag.
Test: ./httpx -silent -cname -ports 80,443 -l list.txt

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1230,7 +1230,7 @@ retry:
 		builder.WriteString(fmt.Sprintf(" [%s]", ip))
 	}
 
-	ips, cnames, err := getDNSData(hp, domain)
+	ips, cnames, err := getDNSData(hp, URL.Host)
 	if err != nil {
 		ips = append(ips, ip)
 	}


runner/runner.go
```diff
@@ -1230,7 +1230,7 @@ retry:
 		builder.WriteString(fmt.Sprintf(" [%s]", ip))
 	}
 
-	ips, cnames, err := getDNSData(hp, domain)
+	ips, cnames, err := getDNSData(hp, URL.Host)
```
Mzack9999, line 1233: I think we should cleanup up the URL within `getDNSData` with something like

```go
func getDNSData(hp *httpx.HTTPX, hostname string) (ips, cnames []string, err error) {
	// cleans up the domain from:
	// Prefixes: http:// and https://
	hostname = stringsutil.TrimPrefixAny(hostname, "http://", "https://")
	// Suffixes: :port/path
	if idx := strings.Index(hostname, ":"); idx >= 0 {
		hostname = hostname[:idx]
	}
	dnsData, err := hp.Dialer.GetDNSData(hostname)
...
```

ehsandeep (dismissed):
As per changes suggested by @Mzack9999

Mzack9999 (approved):

ehsandeep (approved):

PR #646, opened 2022-05-31
updates from urlutil library

LuitelSamikshya:
This PR is dependent on the changes in https://github.com/projectdiscovery/urlutil/pull/6
Test: echo https://example.com/#myfragment/path1/path2/test.html | httpx


Mzack9999 (changes requested):
Depends on https://github.com/projectdiscovery/urlutil/pull/6#pullrequestreview-992666031

ehsandeep (approved):

PR #670, opened 2022-06-21
Adding httpx as a library minimal example

Mzack9999:
## Description
This PR contains updated documentation with a minimal example of how to use `httpx` as a library and an integration test.
The runner API needs a major rework to separate the execution logic from gologger/goflags/CLI dependency. The following changes would be required and will be handled into independent follow-up tickets:
- Input stream via channel
- Adding context support (hence support for cancelation/deadline/timeout)
- Output stream via channel

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,6 +451,45 @@ https://docs.hackerone.com
 https://support.hackerone.com
 ```
 
+### Using httpx as a library
+`httpx` can be used as a library by creating an instance of the `Option` struct and populating it with the same options that would be specified via CLI. Once validated, the struct should be passed to a runner instance (to close at the end of the program) and the `RunEnumeration` method should be called. Here follows a minimal example of how to do it:
+
+```go
+package main
+
+import (
+	"log"
+	"os"
+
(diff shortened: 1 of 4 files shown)


jimen0:
Hi @Mzack9999 - Does this PR mean there will be a major version bump for httpx? This is a public facing API change:

```diff
- func (options *Options) validateOptions() {
+ func (options *Options) ValidateOptions() error {
```

Please, note any wrapped error returned from an exported method automatically becomes part of your API. 

```diff
	if options.matchContentLength, err = stringz.StringToSliceInt(options.OutputMatchContentLength); err != nil {
- 		gologger.Fatal().Msgf("Invalid value for match content length option: %s\n", err)
+ 		return errors.Wrap(err, "Invalid value for match content length option")
```

Do you want `stringz` package errors to become part of `httpx/runner` API? It sounds more like an internal package instead.

Happy to beta-test the new httpx public API once it's ready before a new version is tagged, by the way!

Ice3man543 (approved):

ehsandeep (approved):

PR #677, opened 2022-07-01
modified error message with invalid ports

LuitelSamikshya:
This PR modifies the error and warning message with invalid ports.
Test:

```console
echo google.com | ./httpx -p 8000-7000 -v
echo google.com | ./httpx -p 66000 -v
echo https://google.com:777777777 | ./httpx -v
```

### DIFF
--- a/common/customports/customport.go
+++ b/common/customports/customport.go
@@ -1,6 +1,7 @@
 package customport
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


common/customports/customport.go
```diff
@@ -28,6 +28,39 @@ func (c *CustomPorts) String() string {
 	return "Custom Ports"
 }
 
+// ValidateCustomPorts to validate the custom port range
+func ValidateCustomPorts(value string) {
```
Mzack9999, line 32: I think we should change the signature to `func ValidateCustomPorts(value string) error` and just return the errors here.

runner/options.go
```diff
@@ -510,7 +511,15 @@ func (options *Options) ValidateOptions() error {
 	if len(options.OutputMatchCdn) > 0 || len(options.OutputFilterCdn) > 0 {
 		options.OutputCDN = true
 	}
-
+	// Validate the custom port range
+	if len(options.CustomPorts) > 0 {
+		for _, value := range options.CustomPorts {
+			potentialPorts := sliceutil.Dedupe(strings.Split(value, ","))
+			for _, port := range potentialPorts {
+				customport.ValidateCustomPorts(port)
```
Mzack9999, line 519: here we should check if `customport.ValidateCustomPorts(xx)` returned an error and either propagate/print

Mzack9999 (changes requested):

common/customports/customport.go
LuitelSamikshya: Done.

Mzack9999 (approved):
```console
$ echo scanme.sh | go run . -p 8000-7000 -silent
[FTL] First value of port range should be lower than the last port from your range: [8000, 7000].
```

Issue #712, opened 2022-07-28
Could not cast port to integer


JoshuaMart:
### Current Behavior:
With the 1.2.3 update it is no longer possible to use the `https:443` syntax in the `-port` parameter

```
root@scw-amsnetukjzso:~# echo "example.com" | httpx -ports https:443
[FTL] Could not cast port to integer from your value: https:443. Resulting error: strconv.Atoi: parsing "https:443": invalid syntax.
```

### Expected Behavior:
Same as httpx v1.2.2
```
root@scw-amsnetukjzso:~# echo "example.com" | httpx -ports https:443
https://example.com
```

### Steps To Reproduce:
```
echo "example.com" | httpx -ports https:443
```

I think PR #677 is the cause of this error

ehsandeep:
@JoshuaMart see https://github.com/projectdiscovery/httpx/issues/704, will be fixed soon.

Issue #704, opened 2022-07-20
Ports Flag is not working with scheme


amirnsahmad:
### httpx version: 
1.2.3

### Current Behavior:
 When the ports flag is used within scheme (-p https:443) it is breaking.

### Expected Behavior: 
It should probe https on port 443.

### Steps To Reproduce:

1. Run 'echo example.com | httpx -ports https:443'

### Anything else:
Response:
'[FTL] Could not cast port to integer from your value: https:443. Resulting error: strconv.Atoi: parsing "https:443": invalid syntax.'

jimen0:
This isn't a bug. That's now how the `-ports` flag works. Per the docs:

```text
-p, -ports string[]  ports to probe (nmap syntax: eg 1,2-10,11)
```

ehsandeep:
It looks like a bug introduced with [recent](https://github.com/projectdiscovery/httpx/pull/677) changes @jimen0, as custom scheme input is not valid anymore.

```console
echo example.com | ./httpx -ports https:443

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/              v1.2.3

		projectdiscovery.io

Use with caution. You are responsible for your actions.
Developers assume no liability and are not responsible for any misuse or damage.
[FTL] Could not cast port to integer from your value: https:443. Resulting error: strconv.Atoi: parsing "https:443": invalid syntax.
```

jimen0:
Then the docs should be updated too, `-ports` flag help message doesn't indicate that format is supported. @ehsandeep

PR #713, opened 2022-07-28
Bugfix custom port check

JoshuaMart:
- Fix for #704 & #712
- Add New test case

I tested locally and my change seems to solve the problem, I tested via the following commands : 
```
 echo "example.com" | ./httpx -ports 443
 echo "example.com" | ./httpx -ports 442-443
 echo "example.com" | ./httpx -ports https:443
 echo "example.com" | ./httpx -ports http:80,https:443
```


ehsandeep:
@JoshuaMart thanks for the PR, this is now fixed in https://github.com/projectdiscovery/httpx/pull/714 with some additional improvements.

PR #718, opened 2022-08-04
favicon flag improvement

### LINKED ISSUE #702: Favicon flag improvement
Currently, `-favicon` flag needs to be used separately in the workflow to calculate **favicon** hashes of hosts to probe, as such httpx needs to be executed twice and separately, and that can be **optionally** achieved in the same execution.

This changes nothing in terms of favicon calculation or logic execution but is more of an improvement merging results of 2 execution into one to make the feature more practical + usable.

Current behavior:

```console
echo hackerone.com | httpx -silent -sc -ct
https://hackerone.com [302] [text/html]

echo hackerone.com | httpx -silent -favicon
https://hackerone.com/favicon.ico [595148549]
```

Updated behavior:

```console
echo hackerone.com | httpx -silent -sc -ct -favicon
https://hackerone.com [302] [text/html] [595148549]
```
 
More streamlined workflow for users who are also interested in collecting favicon information along with host probing.

LuitelSamikshya:
close #702

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1345,9 +1345,8 @@ retry:
 	}
 	var faviconMMH3 string
 	if scanopts.Favicon {
-		favURL := fmt.Sprintf("%s://%s/favicon.ico", protocol, URL.Host)
-		favreq, _ := hp.NewRequest(method, favURL)
-		if faviconResp, favErr := hp.Do(favreq, httpx.UnsafeOptions{}); favErr == nil {
+		req.URL.Path = "/favicon.ico"
+		if faviconResp, favErr := hp.Do(req, httpx.UnsafeOptions{}); favErr == nil {
 			faviconMMH3 = fmt.Sprintf("%d", stringz.FaviconHash(faviconResp.Data))
 			builder.WriteString(" [")
 			if !scanopts.OutputWithNoColor {
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -995,16 +995,6 @@ retry:
 	if scanopts.Unsafe {
 		req.Header.Add("Connection", "close")
 	}
-	var faviconResp *httpx.Response
-	if scanopts.Favicon {
-		var favErr error
-		favURL := fmt.Sprintf("%s://%s/favicon.ico", protocol, URL.Host)
-		favreq, _ := hp.NewRequest(method, favURL)
-		faviconResp, favErr = hp.Do(favreq, httpx.UnsafeOptions{})
-		if favErr != nil {
-			return Result{URL: URL.String(), Input: origInput, err: favErr}
-		}
-	}
 	resp, err := hp.Do(req, httpx.UnsafeOptions{URIPath: reqURI})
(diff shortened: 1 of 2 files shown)


runner/runner.go
```diff
@@ -995,6 +995,16 @@ retry:
 	if scanopts.Unsafe {
 		req.Header.Add("Connection", "close")
 	}
+	var faviconResp *httpx.Response
```
Mzack9999, line 998: favicon request should be moved after the [primary probe](https://github.com/projectdiscovery/httpx/blob/78762865305f08a6b2788e8a2040a80a71921139/runner/runner.go#L998). For example it can be moved at: https://github.com/projectdiscovery/httpx/blob/78762865305f08a6b2788e8a2040a80a71921139/runner/runner.go#L1348

runner/runner.go
```diff
@@ -995,6 +995,16 @@ retry:
...
+	if scanopts.Favicon {
+		var favErr error
+		favURL := fmt.Sprintf("%s://%s/favicon.ico", protocol, URL.Host)
+		favreq, _ := hp.NewRequest(method, favURL)
+		faviconResp, favErr = hp.Do(favreq, httpx.UnsafeOptions{})
+		if favErr != nil {
```
Mzack9999, line 1004: Favicon errors are not fatal. Consider only printing a warning in verbose/debug mode and continuing the execution

Mzack9999 (changes requested):

runner/runner.go
LuitelSamikshya: Done

runner/runner.go
LuitelSamikshya: Done

runner/runner.go
```diff
@@ -1343,17 +1343,22 @@ retry:
...
 	}
-
 	var faviconMMH3 string
 	if scanopts.Favicon {
-		faviconMMH3 = fmt.Sprintf("%d", stringz.FaviconHash(resp.Data))
-		builder.WriteString(" [")
-		if !scanopts.OutputWithNoColor {
-			builder.WriteString(aurora.Magenta(faviconMMH3).String())
+		favURL := fmt.Sprintf("%s://%s/favicon.ico", protocol, URL.Host)
+		favreq, _ := hp.NewRequest(method, favURL)
```
Mzack9999, line 1349: I think we can reuse the original request (it's not referenced anymore in the code) and avoid memory allocations:
```go
req.URL.Path = "/favicon.ico"
if faviconResp, favErr := hp.Do(req, httpx.UnsafeOptions{}); favErr == nil {
...
```

Mzack9999 (changes requested):

runner/runner.go
LuitelSamikshya: Done

Mzack9999 (approved):
```console
$ echo 192.168.1.1 | go run . -favicon
...
https://192.168.1.1 [-219752612]
```

ehsandeep (approved):

Issue #737, opened 2022-08-24
Add flag for accepting single or comma separated input values


forgedhallpass:
Currently `httpx` only accepts input through piping or from a file. Adding a new option to accept a single input would be convenient in certain cases (e.g. debugging).

```yaml
INPUT:
   -u, -target string      input target host to probe
```

ariasmn:
Hi! 

Wanted to contribute to this amazing project (that a friend of mine heavily uses :+1: ). Just opened a [PR for this issue](https://github.com/projectdiscovery/httpx/pull/759)
As I commented, there might be a couple of things that need some work. Please let me know in that case!

ariasmn:
Closed with #759

PR #759, opened 2022-09-09
Added flag for single or comma separated stdin input values (#737)

ariasmn:
There are a couple of things which I don't know if are ok:
- Variable/function naming. Non-native speaker so maybe the spelling is wrong.
- You can input from the stdin and a file (using both the **_-l_** and _**-u**_ flags) at the same time.
- Did not add any tests because I didn't fully understand how these works. I can do some if needed with some guidance.


ariasmn:
Hey, just wanted to check the current status of the PR and ask if there is anything more I could do to help

runner/runner.go
```diff
@@ -316,8 +316,19 @@ func (r *Runner) prepareInputPaths() {
 }
 
 func (r *Runner) prepareInput() {
-	// check if file has been provided
 	var numHosts int
+	// check if a string input has been provided
+	if r.options.InputTargetHost != "" {
```
forgedhallpass, line 321: We should use the logic from `goflags` that automatically splits the input into a slice. See `StringSliceVarP`

forgedhallpass: You could check how it has been done in other projects of ours (e.g. `nuclei` or `dnsx`)

forgedhallpass (changes requested):
Please see my comments.

ariasmn:
Done! Just updated the PR.

Please let me know if everything is ok.

Ice3man543 (approved):
lgtm

forgedhallpass (approved):

ehsandeep (approved):
```console
./httpx -u example.com,google.com -silent

https://google.com
https://example.com
```

PR #812, opened 2022-10-13
include dsl tag for all json keys

LuitelSamikshya:


Mzack9999:
@ehsandeep shall we keep the `dsl` specific tag or infer the dsl list directly from the `json` tag since the name is supposed to be the same?

ehsandeep:
> infer the dsl list directly from the json tag since the name is supposed to be the same.

@Mzack9999 this, since the name is supposed to be the same, we can directly list all the existing keys from json instead of maintaining them separately.

Mzack9999 (changes requested):
As per @ehsandeep confirmation, the list should be inferred from the `json` tag:
- [ ] Remove `dsl` tag
- [ ] Build the list using `json` tag and its attributes

runner/types.go
```diff
@@ -24,47 +24,47 @@ func (o AsnResponse) String() string {
 
 // Result of a scan
 type Result struct {
-	Timestamp        time.Time   `json:"timestamp,omitempty" csv:"timestamp" dsl:"timestamp"`
+	Timestamp        time.Time   `json:"timestamp,omitempty" csv:"timestamp"`
 	ASN              interface{} `json:"asn,omitempty" csv:"asn"`
```
Mzack9999, line 28: This should be `*AsnResponse`

runner/types.go
```diff
@@ -75,29 +75,41 @@ type Result struct {
 // function to get dsl variables from result struct
 func dslVariables() []string {
```
Mzack9999, line 76: Since we are using `github.com/mitchellh/mapstructure` to produce the map of values for the expression, can't we use the same library to retrieve the list of available key names?

runner/types.go
```diff
@@ -75,29 +75,41 @@ type Result struct {
...
+				for j := 0; j < types.NumField(); j++ {
+					subField := types.Field(j)
+					if subField.Tag.Get("json") != "" {
+						sanitizeTag(subField.Tag.Get("json"))
+					}
 				}
 			}
 		}
+		if field.Name == "Hashes" {
+			dslVarList = append(dslVarList, "header_md5", "header_mmh3", "header_sha256", "header_simhash", "body_md5", "body_mmh3", "body_sha256", "body_simhash")
```
Mzack9999, line 104: It's not necessary to compare at each iteration, we can add them once before returning `dslVarList`

Mzack9999 (changes requested):
Proposed a different approach

Mzack9999 (dismissed):
suggesting different approach

runner/types.go
LuitelSamikshya: Done.

runner/types.go
LuitelSamikshya: I added this logic to preserve the position of md5, mmh3, sha256 and simhash right after hashes. 
Removed the comparison at each iteration.

runner/types.go
LuitelSamikshya: With the use of `github.com/mitchellh/mapstructure` library the fields in the result struct are being removed if it's empty, so we'll need to pass some random values to all the fields which requires hardcoding the values. Also the other issue is it's not creating a flagged map with key value , the map we get is `map[asn:map[as-country: as-name: as-number: as-range:]` which requires multiple typecasts and iterations.

Mzack9999: Implemented it in this way (and fixed few other issues) since we need an exact mapping between dsl variable list and mapstructure generated variable pairs

Mzack9999 (approved):

ehsandeep (dismissed):
```console
./httpx -ldv

[FTL] interface{} not allowed
```

ehsandeep (approved):

PR #823, opened 2022-10-21
Append to file if output file is already created.

ShubhamRasal:


Mzack9999 (approved):
lgtm - `os.Create` uses by default `0666`

ehsandeep (changes requested):
`APPEND` is required only when the `-resume` option is used; otherwise httpx and other pd tools default to write/overwrite to the given output file.

ehsandeep (approved):

PR #825, opened 2022-10-24
Add OnResult options support httpx as library

### LINKED ISSUE #824: Using httpx as library
it seems that httpx is not currently usable as a library

For example: Naabu or nuclei allow using as library:  https://github.com/projectdiscovery/naabu/issues/432

How to use httpx as library? 
If httpx doesn't support using as library yet, please add Resultcallback on options so that it can be used as library.

lucy9x:
Closes #824


Mzack9999 (approved):
lgtm!

PR #842, opened 2022-10-30
Improved httpx feature - favicon detection.

Mre11i0t:
1. Current workflow
   - Request to url + /favicon.ico
   - Doesn't check if the faviconResp.data is html or it's image
2. Implemented workflow
   - Request to url + /favicon.ico if failed, will scrape root path for link element with rel (icon or shortcut icon or mask-icon or apple-touch-icon)
   - Supports checking if the response content is actually an image
   
Do let me know if I need to make any changes :)


Mre11i0t:
Works locally, before running workflows need to modify httpx.go to original version I believe...

cmd/httpx/httpx.go
```diff
@@ -6,8 +6,8 @@ import (
 	"runtime"
 	"runtime/pprof"
 
+	"github.com/Mre11i0t/httpx-updated-favicon/runner"
```
Mzack9999, line 9: this should be `github.com/projectdiscovery/httpx/runner`

common/stringz/stringz.go
```diff
@@ -107,11 +108,18 @@ func GetInvalidURI(rawURL string) (bool, string) {
...
+	// check if contenttype matches image/x-icon
+	contentType := http.DetectContentType(data)
+	// check content type to be any image
+	if strings.HasPrefix(contentType, "image/") {
+		stdBase64 := base64.StdEncoding.EncodeToString(data)
+		stdBase64 = InsertInto(stdBase64, 76, '\n')
+		hasher := murmur3.New32WithSeed(0)
+		hasher.Write([]byte(stdBase64))
+		return int32(hasher.Sum32())
+	} else {
```
Mzack9999, line 120: Here we can skip the `else` part and directly call `return 0`

runner/runner.go
```diff
@@ -1421,13 +1422,59 @@ retry:
...
 			faviconMMH3 = fmt.Sprintf("%d", stringz.FaviconHash(faviconResp.Data))
-			builder.WriteString(" [")
-			if !scanopts.OutputWithNoColor {
-				builder.WriteString(aurora.Magenta(faviconMMH3).String())
+			if faviconMMH3 == "0" {
+				req.URL.Path = "/"
+				if faviconResp, favErr := hp.Do(req, httpx.UnsafeOptions{}); favErr == nil {
+					parsed, err := html.Parse(strings.NewReader(string(faviconResp.Data)))
+					var f func(*html.Node)
+					f = func(n *html.Node) {
```
Mzack9999, line 1430: For better readability, I think this should be moved into an independent external function that performs the recursive call up to a max nested limit and return the hash to print to the output or error

Mzack9999 (dismissed):
Suggesting some changes, please let me know if you need any help

Mre11i0t:
Made the requested changes, can you please review them @Mzack9999 

I have a small doubt when I run go build . in /cmd/httpx if it's pointing towards projectdiscovery/httpx/runner.. how would automated workflows build using the updated runner.go 🤔🤔

Mre11i0t:
I think I saw my mistakes working on fixing it right now

Mre11i0t:
@ehsandeep I think I corrected lint errors could you reperform tests

Mre11i0t:
Any idea why sonar cloud is failing ?

ehsandeep:
@Mre11i0t please ignore sonarcloud error.

Mre11i0t:
Any update @Mzack9999 ? Made required changes to the code

ehsandeep:
@Mre11i0t @Mzack9999 FYI - https://github.com/projectdiscovery/httpx/issues/860

ehsandeep:
@Mre11i0t 

> This PR will actually fix - #860
> 
> Correct me if I'm wrong.

with this PR -

```console
echo https://static.sidefx.com/images/ | go run . -silent -favicon

https://static.sidefx.com/images/
```

with expected fix:

```console
echo https://static.sidefx.com/images/ | go run . -silent -favicon

https://static.sidefx.com/images/ [1759967229]
```

Mre11i0t:
image

Request path update fixed it req.URL.Path += "/favicon.ico"

Mre11i0t:
Another bug which is 

with this PR -

```echo https://sidefx.com/ | go run . -silent -favicon
https://sidefx.com```

expected fix :

```echo https://sidefx.com/ | go run . -silent -favicon
https://static.sidefx.com/images/favicon.ico [1759967229]```

Will improve this PR to fix this bug too 💯

**Comment by ehsandeep** [comment]
@Mre11i0t nice catch; thank you for working on this improvement.

since we are looking to make **one additional request** for the favicon option, here is a workflow we can try to follow let me know if it makes sense to you.

When `-favicon` option is used -

1. Check if current URI is ending with `.ico`, if yes, we can print `body_mmh3` in CLI output without making any additional request.
2. If it's not `.ico` URI (1st case), we can look into the response of the requested path for element and rel `shortcut/mask/apple-touch` icon
	- link with `.ico` extension will be prioritized if available, if not, any of link from other icons can be requested.
3. If it's not the 1st or 2nd case, `/favicon.ico` path can be appended with probed URI.

response content validation can be applied to all cases.

**Comment by Mre11i0t** [comment]
Understood, my current code is confusing and raw even for me 😅. Will re work the code and work on optimized code following the updated workflow.

Doubts:
1. Will favicon flag print only one favicon hash or all available ones on the website. 
2. Can I enable follow-redirects option to true by default as 301s are destined in few cases of fetching favicons.
3. I noticed my code isn't updating debug req/resp and stats, will I work on that part too. Do I need to update anything else apart debug, stats ?

**Comment by Mre11i0t** [comment]
Implemented the workflow, enabled auto redirect if -favicon flag is enabled

Worked for following testcases

image

**Comment by Mzack9999** [comment]
@Mre11i0t apologies for the delayed reply, I managed to have a look only today to your PR. I'm unable to reproduce the case of `302` redirect with `cloudsek.com`:
```console
$ echo cloudsek.com | go run . -favicon -v                             
...
https://cloudsek.com [https://cloudsek.com/favicon.ico] [1131908660]
```
The redirect is correctly followed, in fact the hash matches, but the output is the one extracted from the page

Mre11i0t:
Hi @Mzack9999, not sure why you're not getting proper output.

I even tried that right now and got the following output, not sure why you're getting wrong url there. It correctly followed the workflow and got it from 2nd try which is by scraping the page. If that failed, it would've got it from https://cloudsek.com/favicon.ico

image

Mre11i0t:
Should I update anything? 👀

Mre11i0t:
Any updates ? 😅

ehsandeep:
@Mre11i0t, this PR will be slightly reworked before we merge this; I will keep you posted.

ehsandeep:
@Mre11i0t, this is reworked as part of https://github.com/projectdiscovery/httpx/pull/906; thank you for opening a PR for this.

PR #853, opened 2022-11-07
Add delay between making httpx requests.

ShubhamRasal:
Add delay flag
`-delay string                      duration between each http request (eg: 200ms, 1s)`

- sleep before creating goroutine to make http request
- eg.  
 ``` cat urls.txt | httpx -delay 2s ``` 
     
  ``` cat urls.txt | httpx -delay 2s -port 80,443,8443 ```

### DIFF
--- a/runner/options.go
+++ b/runner/options.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"golang.org/x/exp/maps"
@@ -164,6 +165,7 @@ type Options struct {
 	Retries                   int
 	Threads                   int
 	Timeout                   int
+	Delay                     string
 	filterRegex               *regexp.Regexp
 	matchRegex                *regexp.Regexp
 	VHost                     bool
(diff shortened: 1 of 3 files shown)


ehsandeep (approved):
```console
$ time echo https://www.hackerone.com | ./httpx -silent -delay 1s -path /test,/test2,/test3

https://www.hackerone.com/test
https://www.hackerone.com/test2
https://www.hackerone.com/test3

real	0m4.775s
user	0m0.053s
sys	0m0.048s

$ time echo https://www.hackerone.com | ./httpx -silent -delay 2s -path /test,/test2,/test3

https://www.hackerone.com/test
https://www.hackerone.com/test2
https://www.hackerone.com/test3

real	0m6.671s
user	0m0.052s
sys	0m0.040s
```

runner/options.go
```diff
@@ -396,6 +398,7 @@ func ParseOptions() *Options {
 		flagSet.BoolVarP(&options.ExcludeCDN, "exclude-cdn", "ec", false, "skip full port scans for CDNs (only checks for 80,443)"),
 		flagSet.IntVar(&options.Retries, "retries", 0, "number of retries"),
 		flagSet.IntVar(&options.Timeout, "timeout", 5, "timeout in seconds"),
+		flagSet.StringVar(&options.Delay, "delay", "", "duration between each http request (eg: 200ms, 1s)"),
```
Mzack9999, line 401: I think we can simplify with `flagSet.DurationVarP`

Mzack9999 (changes requested):

Ice3man543 (approved):

runner/options.go
ShubhamRasal: Updated code as per your suggestion.

PR #885, opened 2022-11-22
adds base64 encoding support (#813)

### LINKED ISSUE #813: Add option to return base64-encoded response-body on json output

Add a `-irrb` option.
```
-irrb, -include-response-base64           include base64-encoded http request/response in JSON output (-json only)
```

Would make possible using json output to save binary unaltered by utf-8 encoding (needed by json)

For example, if you download a favicon through httpx, and try to write content returned by `response-body`, the saved files will be different from original favicon.

Having an option to get base64-encoded body allows such use-case.

tarunKoyalwar:
# Description

- Adds base64 encoding to  request and response in json mode.
 
```sh
   -irrb, -include-response-base64   include base64 encoded http request/response in JSON output (-json only)
```
- Below JSON Fields are base64 encoded
  -  request // entire request
  - raw_header // response headers
  - body. // response body

### Run Below Command to Verify

```sh
echo "projectdiscovery.io/xxx" | ./httpx -irrb -json | jq .
```

closes #813

### DIFF
--- a/runner/options.go
+++ b/runner/options.go
@@ -54,7 +54,7 @@ type scanOptions struct {
 	OutputWithNoColor         bool
 	OutputMethod              bool
 	ResponseInStdout          bool
-	B64ResponseInStdout       bool
+	Base64ResponseInStdout    bool
 	ChainInStdout             bool
 	TLSProbe                  bool
 	CSPProbe                  bool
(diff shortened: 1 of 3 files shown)


runner/options.go
```diff
@@ -54,6 +54,7 @@ type scanOptions struct {
 	OutputWithNoColor         bool
 	OutputMethod              bool
 	ResponseInStdout          bool
+	B64ResponseInStdout       bool
```
Mzack9999, line 57: Let's be verbose: `b64` => `base64`

runner/runner.go
```diff
@@ -1771,3 +1778,8 @@ func normalizeHeaders(headers map[string][]string) map[string]interface{} {
 	}
 	return normalized
 }
+
+// b64 returns base64 of given byte array
+func b64(bin []byte) string {
```
Mzack9999, line 1783: maybe we can move this to `common/stringz/stringz.go`?

Mzack9999 (changes requested):
minor changes suggested

Mzack9999 (approved):
lgtm - Example:
```console
$ echo "projectdiscovery.io/xxx" | go run . -irrb -json -o test.json
$ cat test.json | jq
```

ehsandeep (approved):

PR #996, opened 2023-02-20
Update README with correct help flag

m4lwhere:
This updates the README file with the `--help` flag instead of `-h`. The `-h` flag is used for headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ go install -v github.com/projectdiscovery/httpx/cmd/httpx@latest
 # Usage
 
 ```sh
-httpx -h
+httpx --help
 ```
 
 This will display help for the tool. Here are all the switches it supports.


ehsandeep:
@m4lwhere flags are case sensitive, `-h` is short flag for `-help` and `-H` is short flag for `-header`

```console
httpx -h
```

Works as expected, let me know if you see otherwise!

m4lwhere:
Thanks for the quick response, apologies for any miscommunication!

I am getting a different help window after running the Installation Instructions (https://github.com/projectdiscovery/httpx#installation-instructions):

`go install -v github.com/projectdiscovery/httpx/cmd/httpx@latest`


I get a very different help window, which requires the use of the `--help` switch instead of `-h`. 

## When I try to use the -h flag
```
╭─░▒▓ 💣₭ÂĻĭ💣   /opt ············
╰─ httpx -h
Error: -h option requires 2 arguments
```
## What I get when I run --help instead
This appears to be significantly less features than on the README
```
╭─░▒▓ 💣₭ÂĻĭ💣   /opt ································································· ✔  at 19:20:30  ▓▒░
╰─ httpx --help
                                                     HTTPX 🦋

                                          A next generation HTTP client.

Usage: httpx <URL> [OPTIONS]


 -m, --method METHOD             Request method, such as GET, POST, PUT, PATCH, DELETE, OPTIONS, HEAD.
                                 [Default: GET, or POST if a request body is included]

 -p, --params <NAME VALUE> ...   Query parameters to include in the request URL.

 -c, --content TEXT              Byte content to include in the request body.

 -d, --data <NAME VALUE> ...     Form data to include in the request body.

 -f, --files <NAME FILENAME> ... Form files to include in the request body.

 -j, --json TEXT                 JSON data to include in the request body.

 -h, --headers <NAME VALUE> ...  Include additional HTTP headers in the request.

 --cookies <NAME VALUE> ...      Cookies to include in the request.

 --auth <USER PASS>              Username and password to include in the request. Specify '-' for the password to
                                 use a password prompt. Note that using --verbose/-v will expose the
                                 Authorization header, including the password encoding in a trivially reversible
                                 format.

 --proxies URL                   Send the request via a proxy. Should be the URL giving the proxy address.

 --timeout FLOAT                 Timeout value to use for network operations, such as establishing the
                                 connection, reading some data, etc... [Default: 5.0]

 --follow-redirects              Automatically follow redirects.

 --no-verify                     Disable SSL verification.

 --http2                         Send the request using HTTP/2, if the remote server supports it.

 --download FILE                 Save the response content as a file, rather than displaying it.

 -v, --verbose                   Verbose output. Show request as well as response.

 --help                          Show this message and exit.
```

## Expected results
I expected to see the same very detailed and specific flags as the README at https://github.com/projectdiscovery/httpx#usage

## Additional Info
If it helps, I am running this on Kali WSL. Is there a different set of installation instructions I should try instead? Thanks for the help!

ehsandeep:
@m4lwhere that's because you are running another httpx, for more context - https://github.com/projectdiscovery/httpx/discussions/516

m4lwhere:
Awesome, thanks for the help!

PR #1068, opened 2023-03-30
Issue 1019 dsl var missing

ShubhamRasal:
Fixed bug of all dsl variables not accessiable
```
eg:
❯ echo example.com | ./httpx -mdc "contains(body, 'example')" -silent
https://example.com

❯ echo example.com | ./httpx -fdc "contains(body, 'example')" -silent

```

Note:
- If `-mdc` and `-fdc` flag is passed, it will include body in json output similar to `-irr`.
- As `-mdc` and `-fdc` need body content to check. 
```
❯ echo example.com | ./httpx -mdc "contains(body, 'example')" -silent -json | jq
{
  "timestamp": "2023-03-30T12:18:32.170904523+05:30",
  "hash": {
    ...
   body": "<!doctype html>\n<html>\n<head>\n    <title>Example Domain</title>\n\n    <meta charset=\"utf-8\" />\n    <meta http-equiv=\"Content-type\" conte....
   ....
}
```

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -31,7 +31,6 @@ import (
 	"github.com/projectdiscovery/httpx/common/hashes/jarm"
 	"github.com/projectdiscovery/mapcidr/asn"
 	errorutil "github.com/projectdiscovery/utils/errors"
-	mapsutil "github.com/projectdiscovery/utils/maps"
 
 	"github.com/bluele/gcache"
 	"github.com/logrusorgru/aurora"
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):
```console
$ echo example.com | ./httpx -mdc "contains_all(body, 'example', 'domain')" -silent

https://example.com
```

Mzack9999 (approved):
lgtm

PR #1175, opened 2023-05-17
Add ssl certs to output

kiriknik:
Add ssl certs to output

### DIFF
--- a/common/httpx/option.go
+++ b/common/httpx/option.go
@@ -27,6 +27,7 @@ type Options struct {
 	MaxRedirects         int
 	Unsafe               bool
 	TLSGrab              bool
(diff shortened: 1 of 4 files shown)


ehsandeep (commented):
@kiriknik, how is this different from `-tls-grab` option available in jsonl output? do you have an example / usecase of share?

kiriknik:
Yep, i want to get info about ssl certs in simple output ( without json)
Actually, i dont know, that json output much more verbose that simple)

ehsandeep:
@kiriknik thanks for sharing about the case, we will expose this feature later with a different implementation instead of adding a dedication option for it, closing this PR as this will be implemented later.

Issue #1125, opened 2023-05-01
Httpx -path option with -screenshot don't work


1br00:
There is a bug here, when i use this command: subfinder -d <target> | httpx -silent -path "/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E" (path of vulnerable xss)  -mc 200 -follow-redirects -screenshot

It wont screenshot the path, I got in my screenshot folder "index_screenshot.txt" there is no any png... so did you combied a -path option with -screenshot?

Here is full link: subfinder -d maglaj.ba | httpx -silent -path "/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E" -mc 200 -follow-redirects -screenshot

ehsandeep:
@1br00 it is supposed to work with `-path` option and working for me.

```console
$ echo google.com | httpx -silent -path "/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E" -mc 200 -follow-redirects -screenshot

$ cat output/screenshot/index_screenshot.txt 
/Users/geekboy/Github/nuclei/v2/cmd/nuclei/output/screenshot/google.com/2c60e830c34e274ab5c52072cdd3e393eb016a6e.png https://google.com/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E?id=%22%3E%3Cimg%20src=x%20onerror=alert(1)%3E (404 Not Found)
```

what's the actual error in your case?

RFHACKE:
i got the same issue i used httpx to screenshot host but some screenshot are missing 
These missing screenshot files are listed in index_screenshot but the files are not in dir

TheKalin:
Same for me. hese missing screenshot files are listed in index_screenshot but the files are not in dir

dogancanbakir:
Quick update: I'm able to reproduce the issue mentioned above. The screenshots in `index_screenshot.txt` do not correspond with the files in the `screenshot` directory. Working on it...

PR #1193, opened 2023-05-26
Don't write missing screenshots to index file

dogancanbakir:
The screenshots in index_screenshot.txt do not correspond with the files in the screenshot directory. The error causing this issue is that `ScreenshotWithBody` returns `context deadline exceeded` for the following URLs:

1. https://www.maglaj.ba/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E
2. https://maglaj.ba/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E

And we are initializing the `screenshotPath` regardless of whether there is an error and returning it. Thus, we write it to the `index_screenshot.txt` file. There are two issues, actually:

1. Screenshot mismatch, as mentioned in the beginning
2. context deadline exceeded: we're getting an error when trying to capture a screenshot of the page with `ScreenshotWithBody`- the root cause.

To repro the issue(from #1125):
```sh
subfinder -d maglaj.ba | httpx -silent -path "/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E" -mc 200 -follow-redirects -screenshot
```

I was able to fix the first issue by avoiding the initialization of `screenshotPath` beforehand. I'll further investigate the root cause.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -683,7 +683,7 @@ func (r *Runner) RunEnumeration() {
 				indexData := fmt.Sprintf("%s %s (%d %s)\n", resp.StoredResponsePath, resp.URL, resp.StatusCode, http.StatusText(resp.StatusCode))
 				_, _ = indexFile.WriteString(indexData)
 			}
-			if indexScreenshotFile != nil {
+			if indexScreenshotFile != nil && resp.ScreenshotPath != "" {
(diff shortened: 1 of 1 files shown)


Mzack9999 (approved):
lgtm
```console
$ go run . -silent -path "/ba/investitori.php?id=%22%3E%3Cimg%20src=x%20onerror=alert(1);%3E" -mc 200 -follow-redirects -screenshot -u http://192.168.1.1
```
image

PR #1213, opened 2023-06-06
add option no-stdin

Ciyfly:
When I run httpx with celery, it gets stuck. After locating the problem, I hope to control it by adding the no-stdin parameter, just like naabu. I hope I can help more people.


ehsandeep (approved):

ehsandeep:
Thank you for the PR @Ciyfly

Issue #1208, opened 2023-06-04
Basic HTML Index file to visualize created screenshot 


saurinn:
### Feature request:
Add a folder with all the screenshots there so it should be easier to skim through them. 

### Use case of this feature:
It is really tedious to go back to another folder so we can see the next screenshot, that's why having a folder or at least a HTML with all the resulting images, sort of like [EyeWitness](https://github.com/FortyNorthSecurity/EyeWitness) does, would be so useful in terms of productivity.

MetzinAround:
Thanks for the request! I think part of screenshot capability is that you can designate a folder where they go.  

<img width="874" alt="Screenshot -0605 at 08 15 @2x" src="https://github.com/projectdiscovery/httpx/assets/65838556/1059c67b-d8fb-4c2f-8440-526282062881">

saurinn:
@ehsandeep exactly! Because it doesn't matter where I designate the custom folder with the -srd flag, they are going to be divided on their related dns name key folder. So, that HTML file with all the pictures indexed will be so useful to have.

PR #1226, opened 2023-06-12
Feat screenshot html

bugrakocabay:
As requested [here](https://github.com/projectdiscovery/httpx/issues/1208), I have implemented a feature for -screenshot flag, which saves the result screenshots in an indexed html file as this;

image

The basic functionality is ready, but there is still room for improvement. Open for any suggestions.

- [x] Populate Response Info field with given flags, such as status code, IP, Title etc.


bugrakocabay:
Populated response info column with the flags provided by user
image

review please

Mzack9999 (changes requested):
@bugrakocabay This is a very cool feature! What do you think about moving the html code into an external html template that can be populated with template syntax (either [go native](https://gowebexamples.com/templates/) or something like [twig like syntax](https://github.com/Masterminds/sprig))? In this way the output report would be completely customizable

bugrakocabay:
Spinned up a new goroutine with a channel, as when I used it inside the one above it, some race conditions occurred.

Mzack9999 (approved):
lgtm!
image

ehsandeep (approved):
thank you @bugrakocabay for adding this support.

PR #1264, opened 2023-06-30
fix github.com/bxcodec/faker/v4 is deprecated

anhnmt:
fix github.com/bxcodec/faker/v4 is deprecated: use github.com/go-faker/faker/v4 instead


Mzack9999 (approved):
lgtm - Nice catch!

PR #1316, opened 2023-08-09
Add flag to preview first N characters

### LINKED ISSUE #935: Flag to display first N characters of the response body

It'd be nice to have a flag to show the first N characters of the response body with httpx.

Example usage:

```
$ cat urls.txt | httpx -title -body-preview 100

http://something.com:25 [] [220 environment.something.org ESMTP Postfix (Debian/GNU)221 2.7.0 Error: I can break rules, too. Go]
```

In this case, the response had no HTML tags.
However, maybe it should be considered to strip all HTML tags so that the response preview only contains the useful text.

When monitoring URLs with `httpx -title`, sometimes you stumble upon pages that have no HTML title.
While there are other display flags that can be used to have a sense of what a specific host is running, a preview of the response body would be great.

Thank you

RamanaReddy0M:
closes #935

## test cmd's
```console
✗ go run . -u https://example.com -bp -strip
✗ go run . -u https://example.com -bp=150 -strip=html
✗ go run . -u https://example.com -bp
✗ go run . -u https://example.com -bp=200
```
NOTE: Used dyanmic var to achieve this: https://github.com/projectdiscovery/goflags/pull/130


ehsandeep (changes requested):
We can remove the html tags from response body and only print actual content in the same line, for example:

```console
go run . -u https://example.com -silent -bp=30

https://example.com [Example Dom]
```

Mzack9999 (changes requested):
- [ ] The body should be printed within `... [ body preview ] ...` (maybe we should consider escaping new lines) just like `-title`
- [ ] Consider using https://github.com/microcosm-cc/bluemonday for specific html sanitisation (there should be already an instance at https://github.com/projectdiscovery/httpx/blob/33709e6adbd9a163f1837949569466feed125a4a/common/httpx/httpx.go#L35)
- [ ] Update readme/docs
- [ ] Consider adding the new flags to functional tests

Mzack9999 (approved):
lgtm!

```console
$ echo https://example.com | go run . -bp 
...
https://example.com [<!doctype html>\n<html>\n<head>\n <title>Example Domain</title>\n\n <meta charset="utf-8" />\n <meta]
$ echo https://example.com | go run . -bp -strip
...
https://example.com [ Example Domain This domain is for use in illustrative examples in documents. You may use this domai]
```

ehsandeep (approved):
## CLI

```console
$ echo https://www.hackerone.com | ./httpx -silent -bp
https://www.hackerone.com [Skip to main content 6th Edition of the Hacker Powered Security Report is available for downloadGe]
```

## JSON
```
$ echo https://www.hackerone.com | ./httpx -silent -bp -json | jq -r .body_preview
Skip to main content 6th Edition of the Hacker Powered Security Report is available for downloadGe
```

hktalent:
@RamanaReddy0M 😊
Please support this
-bp=200,500
It means only get
Content with length 500 bytes starting from 200 bytes

This is useful for batch detection of specific vulnerabilities

Of course, regarding the capture of body

PR #1318, opened 2023-08-13
 Adding two new filters && Output optimization

gorgiaxx:
If someone adds '\n' within the HTML title tag, it could lead to abnormal line outputs.

```
https://xxxxx [301] [301
Moved Permanently] [nginx/1.23.1] [111.203.158.238]
```

Using regular expressions for replacement wasn't convenient enough, so I added two filters.
- Response server header
- HTML titlez

Before

```
-fe '(Server: (Tengine|stgw))|(<title>(WAF|Firewall)</title>)'
```

After

```
-fsh Tengine,stgw -ftt WAF,Firewall
```

### DIFF
--- a/runner/options.go
+++ b/runner/options.go
@@ -159,6 +159,8 @@ type Options struct {
 	OutputMatchStatusCode     string
 	OutputMatchContentLength  string
 	OutputFilterStatusCode    string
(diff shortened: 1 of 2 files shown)


gorgiaxx:
i will change this to branch dev

PR #1322, opened 2023-08-15
Two new filters && Output optimization

gorgiaxx:
I previously used an inappropriate branch, and now I've resubmitted the code using the 'dev' branch.

If someone adds '\n' within the HTML title tag, it could lead to abnormal line outputs.

```
https://xxxxx [301] [301
Moved Permanently] [nginx/1.23.1] [111.203.158.238]
```

Using regular expressions for replacement wasn't convenient enough, so I added two filters.
- Response server header
- HTML titlez

Before

```
-fe '(Server: (Tengine|stgw))|(<title>(WAF|Firewall)</title>)'
```

After

```
-fsh Tengine,stgw -ftt WAF,Firewall
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ EXTRACTOR:
 
 FILTERS:
    -fc, -filter-code string            filter response with specified status code (-fc 403,401)
+   -ftt, -filter-title string[]        filter response body with specified html title (-ftt WAF,Firewall)
+   -fsh, -filter-server string[]       filter response with specified server header (-fsh cloudflare,stgw)
    -fep, -filter-error-page            filter response with ML based error page detection
    -fl, -filter-length string          filter response with specified content length (-fl 23,33)
    -flc, -filter-line-count string     filter response body with specified line count (-flc 423,532)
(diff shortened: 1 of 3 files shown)


ehsandeep:
@gorgiaxx, thanks for creating this PR, You have raised a valid issue, but this PR doesn't seem to solve the existing problem, instead of adding a new option, we can try to improve the existing option.

dogancanbakir (changes requested):
@gorgiaxx,
Thank you for your contribution and for bringing this to our attention! As noted by @ehsandeep, we already have options that can handle filtering and matching conditions specified with the `-fdc` and `-mdc` flags, respectively. Instead of adding new options, it would be more helpful to improve the existing `-mdc` and `-fdc` options to address the issues that have been identified. 

Feel free to reach out if you have any questions or need further clarification.

ehsandeep:
@gorgiaxx please feel free to create with host/example so we can work on fixing the existing issue, we are trying to add new options unless it's really needed to avoid the greater number of options this tool has.

PR #1325, opened 2023-08-16
Add Output Match Tech Save

t101804:
Output Match Tech in Options ( work in saving also ) . fully tested


ehsandeep:
@t101804 this is already supported using `-mdc` option:

Example:

```console
echo hackerone.com | httpx -td -mdc 'contains(tech, "Cloudflare")'
```

PR #1338, opened 2023-08-31
Modified storing of redirect response chain

secinto:
Modified storing of redirect response chain to output all redirects and responses.

Previously responses (especially the final get response body) was not stored always (or even never). 
Modified it that if follow redirect is specified not only the request/responses of the redirects but also the final response body is stored.


ehsandeep (commented):
@secinto can you share an example to reproduce this issue?

ehsandeep:
suppressed by https://github.com/projectdiscovery/httpx/pull/1373

PR #1347, opened 2023-09-07
Fixed DNS resolving A records when -ports is set

kz0ltan:
DNS A records were not showing up in the JSON output if -ports was used or the port was added to the target domain name via -u or -l in <domain_name>:<port> format.
The original code tried to use getDNSData with the URL.Host property which in the above cases included the port specification as well, so the DNS query failed, and no A records were added to the output.

This small change fixes that problem.

### Before:

```bash
echo scanme.sh:443 | httpx -json | jq .a
# echo scanme.sh| httpx -ports 443 -json | jq .a

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

		projectdiscovery.io

[INF] Current httpx version v1.3.4 (latest)
null
```

### After:

```bash
echo scanme.sh:443 | ./httpx -ip -json | jq .a

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

		projectdiscovery.io

[INF] Current httpx version v1.3.4 (latest)
[
  "128.199.158.128",
  "2400:6180:0:d0::91:1001"
]
```


ehsandeep (approved):
@kz0ltan thank you for fixing this issue!

Mzack9999 (approved):

ehsandeep (approved):

PR #1431, opened 2023-11-06
fix windows path issue

dogancanbakir:
Closes #1421.

Note: It still needs to be tested. However, I am setting up a Windows test environment on AWS. I'll make sure to follow up with you once testing is complete.

PR #1437, opened 2023-11-07
fix windows path issue

enomothem:
fix windows path issue


dogancanbakir:
duplicate of https://github.com/projectdiscovery/httpx/pull/1431

PR #1461, opened 2023-11-25
Update -tls-probe option to probe extracted TLS dns_names from all targets

### LINKED ISSUE #1460: httpx -tls-probe only appears to work for first target in list.
### httpx version: 

v1.3.7

### Current Behavior: 

Using -tls-probe option does not probe all the extracted dns_names from all targets' TLS certificates. httpx -tls-probe will only run a tls-probe on a the extracted dns_names from the first target in the list and doesn't tls-probe the other targets.  

### Expected Behavior: 

My expectation, based on the httpx tls-probe help output, is that a tls-probe should be generated for every dns_name extracted from the SubjectCN and SubjectANs from the TLS certificate of each individual target.  

### Steps To Reproduce:
1. Run httpx -tls-probe against a list targets that have multiple different dns_names (example list: 8.8.8.8,1.1.1.1)
2. See that httpx will only attempt a tls-probe against the extracted dns_names from the TLS certificate of the first target. (example list: 8.8.8.8.) 

### Anything else:

Example target list:
image

Example target list after fix:
image

CyberStryk:
This change resolves #1460 by removing the `scanopts.TLSProbe=False` lines inside the TLSProbe code sections. This allows TLSProbe to remain true for the rest of the targets in the list. 

Example target list:
8.8.8.8
1.1.1.1 

Example before change:
image

Example after change:
image


dogancanbakir (approved):
LGTM

```console
$ go run . -tls-probe -l scan.lst

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.3.7 (latest)
https://8.8.8.8
https://1.1.1.1
https://dns.google
https://8888.google
https://cloudflare-dns.com
https://dns.google.com
https://one.one.one.one
```

ehsandeep (approved):

PR #1469, opened 2023-11-27
Strictly use resolvers input list when specified

karelorigin:
The current latest build of HTTPX will always rely on local resolvers by getting them from `/etc/resolvers.conf` regardless of whether the `-r` flag is set. In many scenarios, these resolvers are prone to throttling, causing HTTPX to silently discard domains where these rate limits are being enforced. As a result, many URLs will be skipped and the output of the HTTPX tool is never really complete.

This PR fixes these issues by only allowing local resolvers to be used when a resolver list is not specified by the user.


karelorigin:
GitHub Action workflow was failing due to an issue in the base branch. Just merged it into this PR, so feel free to rerun the workflows. It should function properly now.

Mzack9999:
Thanks for the PR @karelorigin - I've forked it as I was unable to push a small comment change upstream. I'm closing this PR and moved the code to https://github.com/projectdiscovery/httpx/pull/1541. Please let me know if anything is missing.

PR #1497, opened 2023-12-21
Remove additional `-U` for `apk` in Dockerfile

PeterDaveHello:
There is no need to use `-U` with `--no-cache` when using `apk` on Alpine Linux, as using `--no-cache` will fetch the index every time and leave no local cache, so the index will always be the latest, without temporary files remain in the image.


Mzack9999 (approved):
nice catch!

PR #1517, opened 2024-01-04
introduce `-exclude` flag

### LINKED ISSUE #1427: Extend exclude pattern functionality
`exclude-cdn` and `exclude-private-ips` should be deprecated, then merged and extended into a unique `exclude` flag that accepts:
- CIDRs (`x.x.x.x/y`)
- Hosts (`*.scanme.*.sh`)
- Regexes (`^test$`)
- CDNs (`cdn`)
- Private IPs (`private-ips`)
- Ports (`*:8080`)

Example:

```console
httpx -exclude cdn,private-ips,http://*scanme.sh,*:8080,1.0.0.0/24
```

Each tools has its own implementation, hence it would be interesting implementing it as a shared logic that could be plugged into all of them. For example we can add a callback to `fastdialer.OnBeforeConnect` where we check all supported patterns.

dogancanbakir:
Closes #1427

```console
go run . -exclude 'cdn,private-ips,http://*scanme.sh,443,1.0.0.0/24' -u http://scanme.sh,127.0.0.1,scanme.sh:443,projectdiscovery.io,example.com -v  

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.3.7 (latest)
[DBG] Skipping private host 127.0.0.1
[DBG] Failed '127.0.0.1': target has a private ip and will only connect within same local network
[DBG] Skipping URL matching exclude regex: http://scanme.sh
[DBG] Failed 'http://scanme.sh': url matches an exclude regex
[DBG] Skipping cdn target: projectdiscovery.io:
[DBG] Failed 'projectdiscovery.io': cdn target only allows ports 80 and 443
https://example.com
[DBG] Skipping excluded port: scanme.sh:443
[DBG] Failed 'scanme.sh:443': port is in the exclude list
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,8 +233,7 @@ OPTIMIZATIONS:
    -nf, -no-fallback                  display both probed protocol (HTTPS and HTTP)
    -nfs, -no-fallback-scheme          probe with protocol scheme specified in input 
    -maxhr, -max-host-error int        max error count per host before skipping remaining path/s (default 30)
-   -ec, -exclude-cdn                  skip full port scans for CDN/WAF (only checks for 80,443)
-   -eph, -exclude-private-hosts       skip any hosts which have a private ip address
+   -e, -exclude string[]              exclude host matching specified filter (comma separated)
    -retries int                       number of retries
    -timeout int                       timeout in seconds (default 10)
    -delay value                       duration between each http request (eg: 200ms, 1s) (default -1ns)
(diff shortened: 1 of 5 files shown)


Mzack9999 (changes requested):
Looks like that most of the logic is already implemented within fastdialer/networkpolicy, it's enough to pass the Allow/Deny list (the only case not handled is ASN expansion). In order to make the CLI more readable maybe we can consider accepting an input file as well between the potential values and consider each line as an exclusion item. What do you think?

Mzack9999 (approved):

PR #1547, opened 2024-01-20
Reducing heap memory

Mzack9999:
`encoding/gob` allocates lots of bytes buffers that are not recycled until a very long delay by `gc` as the encoding process uses reflection, and go has a conservative approach preventing such streams to be recycled.
This PR avoids marshal/unmarshal and reflection-preventing gc by introducing a typed cache within fastdialer.

With a 200k targets list, the in use memory went down from ca 4Gb to ca 30Mb:

```console
$ wc -l ../targets/targets.txt 
  200000 ../targets/targets.txt
$ head ../targets/targets.txt 
https://cmlfqasbsl1ojeov2is0:8000/?a=0
https://cmlfqasbsl1ojeov2isg:8000/?a=1
...
# disabling rate limit to increase pressure on heap
$ go run . -rl -1 -l ../targets/targets.txt -o test.txt -status-code
```

Screenshot 2024-01-20 at 14 49 22

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,7 +45,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/projectdiscovery/asnmap v1.0.6
 	github.com/projectdiscovery/dsl v0.0.37
-	github.com/projectdiscovery/fastdialer v0.0.55-0.20240120134314-6531124bcf55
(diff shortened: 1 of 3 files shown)


Mzack9999:
Tests failing because of unrelated ASN api availability - Merging

PR #1553, opened 2024-01-25
Fix exclude cdn option

### LINKED ISSUE #1417: -ec flag is ignored when -ports is used
### httpx version:
v1.3.5

### Current Behavior:
httpx ignores the `-ec` flag when passing the `-ports` argument

### Expected Behavior:
CDN hosts are excluded

### Steps To Reproduce:
```bash
echo HOST | httpx -ec -threads 300 -no-fallback -random-agent -ports 80,443
```

image

RamanaReddy0M:
- closes #1417 

```console
✗ go run . -u projectdiscovery.io,scanme.sh -cdn -e cdn

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.3.9 (latest)
https://scanme.sh
```

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -2267,7 +2268,7 @@ func (r *Runner) skipCDNPort(host string, port string) bool {
 		return false
 	}
 
-	if isCdnIP && sliceutil.Contains(r.options.CustomPorts, port) {
+	if isCdnIP && slices.Contains(r.options.CustomPorts, port) {
 		return true
 	}
 	// If the target is part of the CDN ips range - only ports 80 and 443 are allowed
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -2267,11 +2267,11 @@ func (r *Runner) skipCDNPort(host string, port string) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


runner/runner.go
```diff
@@ -2267,6 +2267,9 @@ func (r *Runner) skipCDNPort(host string, port string) bool {
 		return false
 	}
 
+	if sliceutil.Contains(r.options.Exclude, "cdn") && isCdnIP {
```
dogancanbakir, line 2270: we can use `r.scanopts.ExcludeCDN` instead of `sliceutil.Contains(r.options.Exclude, "cdn")`. But I think there is no need for that since we're checking at the beginning of this func.

RamanaReddy0M: > Also, we're changing the behavior of this function by only checking against `isCdnIP`. Intented?

Previously observed that the logic isn't working without custom ports. But this seems conflicting, Updated the logic to check on custom ports with existing behaviour. can you test now?
cc: @dogancanbakir

Mzack9999 (approved):
lgtm - As a follow up I think we can also create an issue in [utils](https://github.com/projectdiscovery/utils) to use standard library calls (eg. `slices.Contains(..)` from go 1.21 or just use them directly. They had meaning when they were missing from the language standard library.

dogancanbakir (approved):

PR #1608, opened 2024-03-02
Introducing CanHaveTitleTag Function for MIME Type Validation

RedYetiDev:
This pull request introduces a new function called `CanHaveTitleTag`. The purpose of this function is to validate whether a given MIME type supports the extraction of a `<title>` tag from its content. The motivation behind this addition is to address issues where the extractor incorrectly assumes support for a `<title>` tag, even when it's not applicable, such as in `application/json` MIME types.

**Changes:**

1. Added a new function `CanHaveTitleTag` in `title.go` to perform validation for MIME types supporting `<title>` tag extraction.
2. Updated `runner.go` to use the new function

> [!NOTE]
> I'm not a frequent **go** developer, and this may be an ineffective way to achieve this goal, but from what I can tell, it gets the job done.


RedYetiDev:
Hi! I noticed that some of the builds were failing with `use of untyped nil in assignment`, does that have to do with `title := nil`? How do I fix that?

common/httpx/title.go
```diff
@@ -40,6 +40,25 @@ func ExtractTitle(r *Response) (title string) {
 	return title
 }
 
+func CanHaveTitleTag(mimeType string) bool {
+    mimeTypes := []string{
```
dogancanbakir, line 44: Let's move mime types to the beginning of this file as
```go
var (
	cutset                  = "\n\t\v\f\r"
	reTitle                 = regexp.MustCompile(`(?im)<\s*title.*>(.*?)<\s*/\s*title>`)
	reContentType           = regexp.MustCompile(`(?im)\s*charset="(.*?)"|charset=(.*?)"\s*`)
	supportedTitleMimeTypes = []string{
		"text/html",
		"application/xhtml+xml",
		"application/xml",
		"application/rss+xml",
		"application/atom+xml",
		"application/xhtml+xml",
		"application/vnd.wap.xhtml+xml",
	}
)
```
and refactor this function as

```go
func CanHaveTitleTag(mimeType string) bool {
	return slices.Contains(supportedTitleMimeTypes, mimeType)
}
```

runner/runner.go
```diff
@@ -1589,8 +1589,11 @@ retry:
 		}
 		builder.WriteRune(']')
 	}
+	title := nil
```
dogancanbakir, line 1592: Let's change this section as

```go
var title string
	if httpx.CanHaveTitleTag(resp.GetHeaderPart("Content-Type", ";")) {
		title = httpx.ExtractTitle(resp)
	}

	if scanopts.OutputTitle && title != "" {
		builder.WriteString(" [")
		if !scanopts.OutputWithNoColor {
			builder.WriteString(aurora.Cyan(title).String())
		} else {
			builder.WriteString(title)
		}
		builder.WriteRune(']')
	}
```

dogancanbakir (changes requested):
I left some comments for you; thanks for the PR!

common/httpx/title.go
```diff
@@ -40,6 +49,10 @@ func ExtractTitle(r *Response) (title string) {
 	return title
 }
 
+func CanHaveTitleTag(mimeType string) bool {  
+    return slices.Contains(supportedTitleMimeTypes, mimeType)
```
dogancanbakir, line 53: import `"slices"`

dogancanbakir (changes requested):

common/httpx/title.go
RedYetiDev: Right! Sorry!

RedYetiDev:
I think I undid the dev changes in my push, once this is ready, I'll squash and open a new pull if you'd like.

dogancanbakir (approved):

RedYetiDev:
The latest commit shouldn't cause any issues, I updated the patch to match the code on the dev branch rather than the main branch.

RedYetiDev:
Hi, seeing as there are 11 commits for this, I feel like squashing might be wise, is that done during the merge process, or should I do it and re-open a pull?

RedYetiDev:
Hi, any update on this?

Mzack9999 (approved):

ehsandeep (approved):

RedYetiDev:
Thanks!

PR #1671, opened 2024-04-21
Added the Functionality of storing raw request response headers

saeed0x1:
I have added 2 new flags `-sh` and `-shd`. This functionality works just like `-sr` and `-srd` but It only stores the request and response headers and does not include the response body. 

I was looking for a functionality same as [this](https://github.com/projectdiscovery/httpx/discussions/1215); So I implemented it, Hope it helps.


GeorginaReeder:
Thanks so much for your contribution @saeed0x1 , we appreciate it!

We also have a [Discord server](https://discord.com/invite/projectdiscovery), which you’re more than welcome to join. It's a great place to connect with fellow contributors and stay updated with the latest developments!

ehsandeep (changes requested):
@saeed0x1 thanks for making this pull request.

Instead of adding 2 new options, we can add 1 new option that can be applied to existing output options.

New option that can be added:

```
   -ob, -omit-body  omit response body from output
```

Add this option will be applicable to:

```
   -oa, -output-all                    filename to write output results in all formats
   -sr, -store-response                store http response to output directory
   -csv                                store output in csv format
   -j, -json                           store output in JSONL(ines) format
```

cc @dogancanbakir

ehsandeep (approved):

PR #1699, opened 2024-05-14
feat: updated wappalyzer-go + misc additions

Ice3man543:


runner/runner.go
```diff
@@ -2077,6 +2090,7 @@ retry:
 			"PageType": r.errorPageClassifier.Classify(respData),
 			"pHash":    pHash,
 		},
+		TechnologyDetails: technologyDetails,
```
dogancanbakir, line 2093: How about moving this under "KnowledgeBase" as "Technologies"?

dogancanbakir (approved):

Mzack9999 (approved):
lgtm

runner/runner.go
Ice3man543: Actually, we only want this to be available from sdk. so we can't add it there

PR #1731, opened 2024-05-29
Exclude tech results by default in cli output

### LINKED ISSUE #1730: Remove tech-detect info from default output to stdout
Starting in either 1.6.1 or 1.6.2 the default output to stdout includes the info from --tech-detect. While this info is useful in certain scenarios, it goes against the streamlined integration with other tools/utilities that httpx and other projectdiscovery tools are known for. Now it's required to add the -tech-detect=false flag every time you use httpx and want to pipe the output to another tool or just copy the URL list without extra data. 

This feature request is requesting that the default output be changed back to how it was in 1.6.0 and prior, removing the tech-detect info from stdout by default and only including it when the --tech-detect flag is used. 

Here's a screenshot of an example of the behaviour pre and post upgrade
httpx

I know it's a minor annoyance, but when trying to integrate httpx with other tools and workflows it makes much more sense to include the additional information only when the user requests it rather than by default.

RamanaReddy0M:
- closes #1730

### DIFF
--- a/cmd/integration-test/library.go
+++ b/cmd/integration-test/library.go
@@ -74,7 +74,7 @@ func (h *httpxLibraryWithStream) Execute() error {
 		RateLimit:  150,
 		Retries:    2,
 		Timeout:    10,
-		TechDetect: true,
+		TechDetect: "true",
 		Stream:     true,
 		SkipDedupe: true,
 		OnResult: func(r runner.Result) {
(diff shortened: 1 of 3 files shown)


dogancanbakir:
I don't think we should use dynamicvarp here. We removed it in https://github.com/projectdiscovery/httpx/pull/1702. I also made a PR https://github.com/projectdiscovery/httpx/pull/1732 -missed unassignment lol

Issue #1859, opened 2024-08-07
ztls not work


iuliu8899:
### Current Behavior:
![image](https://github.com/user-attachments/assets/f0877dd6-cb5f-4bc5-9583-79e54c24d1d1)

### Expected Behavior:

it seems the r.Request.Context() done when DialTLS ing. 
https://github.com/projectdiscovery/httpx/blob/5f8e5d0dc5df536fec7cc6189a3e54a6d3a9e60e/common/httpx/tls.go#L65-L68

### Steps To Reproduce:

```
httpx -u google.com -ztls -tls-grab -json -silent | jq .tls
```

### Anything else:

btw, TLSGrab func (not ZTLSGrab) does not set Serial number of the cert...

ehsandeep:
This is now fixed in latest release - https://github.com/projectdiscovery/httpx/releases/tag/v1.6.8

PR #1860, opened 2024-08-08
fix `ztls` context deadline error

RamanaReddy0M:
- closes: #1859

### DIFF
--- a/common/httpx/tls.go
+++ b/common/httpx/tls.go
@@ -63,7 +63,7 @@ func (h *HTTPX) ZTLSGrab(r *http.Response) *clients.Response {
 	}
 	// canonical net concatenation
 	host = net.JoinHostPort(hostname, fmt.Sprint(port))
-	ctx, cancel := context.WithTimeout(context.Background(), h.client2.Timeout)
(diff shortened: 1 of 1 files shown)


dogancanbakir (approved):

common/httpx/tls.go
```diff
@@ -62,7 +63,9 @@ func (h *HTTPX) ZTLSGrab(r *http.Response) *clients.Response {
 	}
 	// canonical net concatenation
 	host = net.JoinHostPort(hostname, fmt.Sprint(port))
-	tlsConn, err := h.Dialer.DialTLS(r.Request.Context(), "tcp", host)
+	ctx, cancel := context.WithTimeout(context.Background(), h.client2.Timeout)
```
Mzack9999, line 66: Shall we use `h.Options.Timeout`?

Mzack9999 (changes requested):

common/httpx/tls.go
RamanaReddy0M: yes. it's default is 10s same as h.client2.Timeout.

Mzack9999 (approved):

PR #1871, opened 2024-08-13
add disable stdout option

### LINKED ISSUE #1692: Changing runner.Options doesn't change live domain output, only goes to stdout
### httpx version:
v1.6.0

### Current Behavior:
When running a scan with httpx, it outputs the domains to stdout. It doesn't seem to be possible to put the live domains into a map, leaving the terminal empty.

`ResponseInStdout` [exists in the options.go](https://github.com/projectdiscovery/httpx/blob/76e61cc6036ec6533460777a355ffc8557eba77b/runner/options.go#L211), but changing/setting this in `runner.Options` doesn't seem to make a difference.

### Expected Behavior:
When changing `runner.Options`, it should be possible to change where live domains are sent (i.e. to a map, instead of to stdout).

### Steps To Reproduce:
`runner.Options` settings look like this, and seem to return an error:

Screenshot 2024-05-08 at 13 06 42

### Anything else:
Reported in [Discord here](https://discord.com/channels/695645237418131507/716605549923270717/1237403503115112539).

dogancanbakir:
Closes #1692

```go
package main

import (
	"fmt"
	"log"

	"github.com/projectdiscovery/goflags"
	"github.com/projectdiscovery/httpx/runner"
)

func main() {
	results := map[string]struct{}{}

	options := runner.Options{
		Methods:         "GET",
		Timeout:         5,
		Threads:         80,
		Silent:          true,
		RandomAgent:     true,
		Retries:         2,
		DisableStdout:   true,
		InputTargetHost: goflags.StringSlice{"scanme.sh", "projectdiscovery.io", "localhost", "x"},
		OnResult: func(r runner.Result) {
			if r.Err != nil {
				return
			}
			results[r.URL] = struct{}{}
		},
	}

	if err := options.ValidateOptions(); err != nil {
		log.Fatal(err)
	}

	httpxRunner, err := runner.New(&options)
	if err != nil {
		log.Fatal(err)
	}
	defer httpxRunner.Close()

	httpxRunner.RunEnumeration()
	fmt.Println("Results:", results)
}
```

```console
$ go run .
Results: map[https://projectdiscovery.io:{} https://scanme.sh:{}]
```

### DIFF
--- a/runner/options.go
+++ b/runner/options.go
@@ -309,6 +309,7 @@ type Options struct {
 	HeadlessOptionalArguments goflags.StringSlice
 	Protocol                  string
 	OutputFilterErrorPagePath string
+	DisableStdout             bool
 	// AssetUpload
 	AssetUpload bool
 	// AssetName
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1098,7 +1098,7 @@ func (r *Runner) RunEnumeration() {
 				}
 			}
 
-			if !jsonOrCsv || jsonAndCsv || r.options.OutputAll {
+			if !r.options.DisableStdout && (!jsonOrCsv || jsonAndCsv || r.options.OutputAll) {
 				gologger.Silent().Msgf("%s\n", resp.str)
 			}
 


Mzack9999 (approved):
lgtm - I guess the tests is failing due to issues with ASN endpoint issues

ehsandeep (approved):

PR #1917, opened 2024-09-15
Added Idle Time Before Screenshot Capture

michael2to3:
Previously, the screenshot feature would capture only the loading screen. Now, you can specify how long to wait before capturing a screenshot, allowing the page to fully load. I've attached two comparison images: the "before" screenshot shows just the loading spinner, while the "after" screenshot displays the fully loaded page
Before:

After:


michael2to3:
That no my fault ;)

Mzack9999 (approved):

PR #1944, opened 2024-10-11
Exclude response, request data from result

### LINKED ISSUE #1914: CSV output ignores command line options for what to extract when use '-filter-condition' argument
### httpx version:
v1.6.8 linux amd64

### Current Behavior:
Output contains response body.

### Expected Behavior:
CSV output doesn't contain response body.

### Steps To Reproduce:
```
httpx -u https://www.baidu.com -csv -o test.csv -sc -title -filter-condition "contains(title,'google')"
```

### Anything else:
https://github.com/projectdiscovery/httpx/issues/1593

dogancanbakir:
Closes #1914

### DIFF
--- a/runner/types.go
+++ b/runner/types.go
@@ -53,7 +53,7 @@ type Result struct {
 	Scheme             string                        `json:"scheme,omitempty" csv:"scheme" mapstructure:"scheme"`
 	Error              string                        `json:"error,omitempty" csv:"error" mapstructure:"error"`
 	WebServer          string                        `json:"webserver,omitempty" csv:"webserver" mapstructure:"webserver"`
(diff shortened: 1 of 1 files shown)


Mzack9999 (changes requested):
What do you think about using something like `github.com/jszwec/csvutil` with `csv:"body,omitempty"`?

dogancanbakir:
Let's create a follow-up issue to assess this option, as it may involve replacing CSV-related mappings with this package.

Mzack9999 (approved):
Follow up ticket at https://github.com/projectdiscovery/httpx/issues/1984 - This PR excludes the fields in export, potentially creating missing data if expected

ehsandeep (approved):

PR #1951, opened 2024-10-14
filter out near duplicate responses

### LINKED ISSUE #758: Add feature to exclude the same http response
while scanning wide scope there are a lot of urls have the same base code so they have the same http response so it's time to let httpx exclude these urls based on comparing every http response with others

dogancanbakir:
Closes #758

```console
$ go run . -u https://httpbin.org -path xml,json,html,html -fd -v

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.6.8 (latest)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
https://httpbin.org/html
[WRN] Skipping duplicate response with simhash 9850149872654300798
https://httpbin.org/xml
https://httpbin.org/json
```


Mzack9999 (changes requested):
- Shall we use something like [gcache](https://github.com/bluele/gcache) with a maximum number of items to avoid memory leak in case of huge input lists?

Mzack9999 (approved):
implementation: lgtm
follow up: update [docs](https://github.com/projectdiscovery/docs/tree/main/tools/httpx) making it clear that only the first response will be considered valid while the others filtered out

ehsandeep (approved):

PR #2059, opened 2025-01-05
Fixed issue with screenshot options when using httpx as a library (ScreenshotIdle option)

v1stra:
`ScreenshotIdle` was not being set when configured in the options type and passed like `h.client, _ = runner.New(&options)`. ScreenshotIdle then defaulted to 0 when passed to `r.browser.ScreenshotWithBody` resulting in no screenshot.


GeorginaReeder:
Thanks for your contribution @v1stra !

Mzack9999 (approved):

PR #2131, opened 2025-03-27
feat(headless): add options to specific if screenshot with full page

yoshino-s:
When screenshot some long page, like phpinfo, it may takes so many resources, but what we need is only a screenshot to prove the result.

So add a new options named `--screenshot-full-page`, with default value as `true` to keep it works as before, to control the rod wheather screenshot it with full page.


Mzack9999 (approved):
lgtm - reversed the logic to allow `-flag` instead of `-flag=false`

Issue #2047, opened 2024-12-20
-follow-redirects should not print location


nadirzb:
### Current Behavior:

Option `-fr` or `-follow-redirects` should not print location, `-location` already does it.

### Expected Behavior:

Option `-fr` or `-follow-redirects` should only follow redirect, not print anything.

### Steps To Reproduce:

echo "redirecteddomain.com" | httpx -fr

ehsandeep:
```
$ echo https://resources.hackerone.com | httpx -silent -sc -fr
https://resources.hackerone.com [301,301,404] [http://resources.hackerone.com/read/account_titles/2763037]
```

Currently redirected URL is printed when `-fr` option is used, expected behavior would be print redirected location only when `-location` option is used.

ehsandeep:
this is fixed!

PR #2173, opened 2025-05-09
fix(runner): don't print out location if not explicitly set as option…

loresuso:
… on follow-redirects

Closes: https://github.com/projectdiscovery/httpx/issues/2047

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL output behavior so that the final URL in the response chain is only included when location output is enabled in the scan options.


ehsandeep (approved):

ehsandeep:
thank you @loresuso

Issue #2218, opened 2025-07-22
Issues with JSON Output in httpx (File Not Generated & Formatting Errors)


ayadim:
Description:

Hi httpx team,

I’m reaching out to report two issues with the JSON output when using the -j option in httpx (v1.7.1):

    The output file is not being generated when I use the -j flag.

    The JSON output (if it does generate) is not properly wrapped in an array ([]), and there are no commas separating individual JSON objects, which makes the output invalid.

```
─$ httpx -l httpx_probed.txt -title -sc -cl -ct -td -cname -ip -j httpx_all_data.json

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.7.1 (latest)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
{"timestamp":"2025-07-22T18:15:56.620296056+01:00","port":"443","url":"https://s1.example.com/","input":"https://s1.example.com/","title":"XXX Tool","scheme":"https","webserver":"nginx/1.18.0 (Ubuntu)","content_type":"text/html","method":"GET","host":"44.X.X.X","path":"/","time":"700.76442ms","a":["44.X.X.X"],"tech":["Bootstrap","Nginx:1.18.0","Ubuntu"],"words":95,"lines":31,"status_code":200,"content_length":1122,"failed":false,"knowledgebase":{"PageType":"other","pHash":0},"resolvers":["1.0.0.1:53","8.8.8.8:53"]}
[...]

PR #2228, opened 2025-08-04
feat: implement JSON export functionality

khs-alt:
This PR addresses issue #2218 

- Add JSON export option with -je flag

`go run cmd/httpx/httpx.go -h`
```
OUTPUT:
...
   -fepp, -filter-error-page-path string  path to store filtered error pages (default "filtered_error_page.json")
   -je, -json-export string               file to export results in JSON format

   -pdu, -dashboard-upload string  upload httpx output file (jsonl) in projectdiscovery cloud (pdcp) UI dashboard
...
```

`go run cmd/httpx/httpx.go -u google.com  -je test.json`
```

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.7.1 (latest)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
https://google.com
 % cat test.json 
[{"timestamp":"2025-08-04T18:24:14.956147+09:00","cdn_name":"google","cdn_type":"cdn","port":"443","url":"https://google.com","input":"google.com","location":"https://www.google.com/","title":"301 Moved","scheme":"https","webserver":"gws","content_type":"text/html","method":"GET","host":"142.250.76.142","path":"/","time":"247.113875ms","a":["142.250.76.142"],"aaaa":["2404:6800:400a:805::200e"],"words":9,"lines":6,"status_code":301,"content_length":220,"failed":false,"cdn":true,"knowledgebase":{"PageType":"other","pHash":0},"resolvers":["1.1.1.1:53","1.0.0.1:53"]}]%  
```

## Summary by CodeRabbit

* **New Features**
  * Export enumeration results to a JSON file.
  * New CLI option --json-export (-je) to specify the JSON output filename.
  * Results are collected during a run and written to the specified JSON file upon completion; export errors are logged.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -780,8 +780,8 @@ func (r *Runner) RunEnumeration() {
 			}
 		}()
 
-		var plainFile, jsonFile, csvFile, indexFile, indexScreenshotFile, jsonExportFile *os.File
-		var jsonExportResults []Result
+		var plainFile, jsonFile, csvFile, indexFile, indexScreenshotFile *os.File
+		var jsonExporter *JSONExporter
 
 		if r.options.Output != "" && r.options.OutputAll {
 			plainFile = openOrCreateFile(r.options.Resume, r.options.Output)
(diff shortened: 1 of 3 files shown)


runner/runner.go
```diff
@@ -1172,6 +1173,31 @@ func (r *Runner) RunEnumeration() {
 				}
 			}
 
+			if r.options.JSONExport != "" {
+				filename := r.options.JSONExport
```
dogancanbakir, line 1177: We can pass `r.options.JSONExport` instead of introducing new var.

runner/runner.go
```diff
@@ -1172,6 +1173,31 @@ func (r *Runner) RunEnumeration() {
...
+
+				if jsonExportResults == nil {
+					jsonExportResults = make([]Result, 0)
+				}
+				jsonExportResults = append(jsonExportResults, resp)
```
dogancanbakir, line 1182: What do you think of copying [this](https://github.com/projectdiscovery/nuclei/blob/dev/pkg/reporting/exporters/jsonexporter/jsonexporter.go) and adding batch support to be nice on memory and make it easier/flexible to play with the implementation details later on. Since the relatively large data could cause memory issues, and we will make sure most of the data is written to the file.

dogancanbakir (changes requested):
Let me know what you think!

runner/runner.go
khs-alt: Thank you for your suggestion, it looks correct!
httpx repo doesn't have a pkg folder, so should I make the same as nuclei `pkg/reporting/exporters/jsonexporter/jsonexporter.go` and add to use this code?
and is it okay to custom `jsonexporter` for using httpx? current `jsonexporter`

runner/runner.go
khs-alt: Thanks for the feedback! I'll update it.

runner/runner.go
dogancanbakir: re 1: let's not go with structural change for now, we can just introduce a new go file
re 2: we can adapt it for the httpx

khs-alt:
Hi! @dogancanbakir  I'm changing the code and the readme file.

And also, I'm trying to check four tags `-irh ` `irr` `irrb` `include-chain` which is written `-json only` also works with using our `-je` tag

Mzack9999:
Thanks for taking the time to work on this! After reviewing the changes and the discussion in [#2218](https://github.com/projectdiscovery/httpx/issues/2218#issuecomment-3191577885)
, we’ve decided not to move forward with this approach. The proposed changes introduce additional complexity and memory overhead without fully addressing the concerns raised, and we’d like to explore a more robust solution in the future.

We appreciate your effort and contribution, but we’ll go ahead and close this PR for now. Thanks again for your time and interest in improving httpx!

Issue #1078, opened 2023-04-08
round robin system for 429 status code


iamNynan:
### Please describe your feature request:

some sort of flag to activate a round-robin system for 429 responses, eg `-rr 4` would mean any request that returned a 429 rate-limiting response would be re-run at the end of the scan after waiting a little bit, the number would dictate how many times it tried to re-run those requests if it kept returning 429.

### Describe the use case of this feature:

rate-limiting can occur in small amounts to where it would not be necessary to stop the scan entirely, but the overall quality of the output would be increased if the URLs could be re-run at a later stage. This is especially common when you use file lists of different sites, or you use multiple `-path`s

Mzack9999:
Thanks for the suggestion. After reviewing this, I think the current feature set already covers the underlying need here, and adding built-in 429 round-robin retry would add complexity that doesn't align well with how httpx works.

**httpx is a one-shot prober** — it takes a list of targets, probes each one, and reports the response. A 429 is a valid, informative response (the host exists and is actively rate-limiting), so recording it is often the right behavior.

**Existing flags that address the root cause:**

- `-rate-limit` / `-rl` — cap requests per second globally
- `-rate-limit-minute` / `-rlm` — cap requests per minute
- `-threads` / `-t` — reduce concurrency to avoid triggering rate limits
- `-retries` — retry failed requests (covers transient errors)

These let you tune the scan to avoid 429s in the first place rather than react to them after the fact.

**For cases where some 429s still slip through**, a two-pass approach works well without any tool changes:

```bash
# first pass: collect results, note 429s
httpx -l targets.txt -fc 429 -o results.txt
httpx -l targets.txt -mc 429 -o retry.txt

# second pass: retry at a lower rate
httpx -l retry.txt -rl 5 -o retry_results.txt
```

This gives you full control over the delay, rate, and number of retries without adding internal state management to httpx.

Going to close this one out, but feel free to reopen if there's a use case the above doesn't cover.

PR #2244, opened 2025-08-21
feat: retry on 429

jjhwan-h:
## Summary

This PR introduces a round-robin retry mechanism for HTTP 429 (Too Many Requests) responses.
Users can control the number of retries and the delay between them using two new flags:

--retry-rounds

--retry-delay

### Options

Added RetryRounds (int) → number of retries for 429 responses

Added RetryDelay (int) → delay between retries (in ms)

Added validation to ensure retries and delay values are valid

### Runner

Integrated a new retryLoop with drainedCh synchronization

Introduced retryJob struct for retry queue management

On 429 responses, jobs are pushed into the retryCh for delayed execution

Used atomic.Int64 to track remaining retries and close drainedCh when done

### Tests

Added TestRunner_Process_And_RetryLoop

srv1: Always returns 429 → retries exhausted, expected 3×429, 0×200

srv2: Returns 2×429 then 200 → expected 2×429, 1×200

Verified drain and retry logic

### Example
httpx -l urls.txt --retry-rounds 2 --retry-delay 500

A request returning 429 will be retried up to 2 times

Each retry happens after 500ms delay

If retries are exhausted, the request is considered failed

## Related Issue
#1078

## Summary by CodeRabbit

* **New Features**
  * Added HTTP 429 retry handling with configurable CLI flags: `--retry-rounds`, `--retry-delay`, and `--retry-timeout`.

* **Documentation**
  * Updated README with guidance on retry flags and Retry-After header behavior.

* **Tests**
  * Added comprehensive test coverage for retry scenarios, including Retry-After header handling and timeout validation.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -23,7 +23,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"golang.org/x/exp/maps"
@@ -1502,15 +1501,7 @@ func (r *Runner) RunEnumeration() {
 	}(nextStep)
 
 	wg, _ := syncutil.New(syncutil.WithSize(r.options.Threads))
-	retryCh := make(chan retryJob)
-
-	retryCtx := context.Background()
-	var retryCancel context.CancelFunc
-	if r.options.RetryRounds > 0 && r.options.RetryTimeout > 0 {
-		retryCtx, retryCancel = context.WithTimeout(retryCtx, time.Duration(r.options.RetryTimeout)*time.Second)
-		defer retryCancel()
-	}
-	_, drainedCh := r.retryLoop(retryCtx, retryCh, output, r.analyze)
+	var rq retryQueue
 
 	processItem := func(k string) error {
 		select {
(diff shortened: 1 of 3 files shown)


Mzack9999:
Nice one! I was thinking about something similar with automatic detection of rate limit from RL headers combined with https://github.com/projectdiscovery/ratelimit/tree/feat-per-key-limit to make the tool compliant with them without slowness.
Just curious, were you testing many different paths towards the same target in order to hit 429? Generally `httpx` use case is to send few requests to a very large number of targets.

jjhwan-h:
Thanks for the feedback! 

In my testing I was indeed sending many different paths to the same target, which resulted in 429 responses. However, in real-world environments with CDNs and WAFs, 429s can also occur sporadically across multiple targets, even with only a few requests per host.  

The `--retry-rounds` flag is therefore intended as an optional safeguard to preserve result accuracy in such cases. It does not alter the default behavior and can be enabled only when needed.  

if you feel additional tests or adjustments to the implementation would make this more useful or easier to maintain, I’d be happy to revise the PR accordingly!

PR #2293, opened 2025-10-16
Strip ANSI color codes from output file while keeping console colors

ayanrajpoot10:
This PR adds support for **removing ANSI color codes from output files** while still allowing colored output in the console.

* Users now get **clean output files** without "garbage" ANSI codes, improving readability.
* Console output retains colors, so users can still see colorized results in real time.
* Before, users had to use the -nc flag to remove ANSI codes, but this also turned off console colors. There was no way to remove the codes while keeping colored console output.
* Minimal performance impact: users with `-nc` see zero impact, and stripping ANSI codes adds only a small cost for colored lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Plain-text outputs no longer include stray ANSI escape sequences when color is disabled, improving readability across primary output paths.
  * Colored output remains unchanged when enabled; JSON and CSV outputs are unaffected.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1183,7 +1183,7 @@ func (r *Runner) RunEnumeration() {
 
 			//nolint:errcheck // this method needs a small refactor to reduce complexity
 			if plainFile != nil {
-				plainFile.WriteString(stripANSI(resp.str) + "\n")
+				plainFile.WriteString(handleStripAnsiCharacters(resp.str, r.options.NoColor) + "\n")
 			}
 
 			if len(r.options.ExcludeOutputFields) > 0 {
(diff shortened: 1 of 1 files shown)


ayanrajpoot10:
@Mzack9999 
If we keep this if condition:

```
if r.options.NoColor {
    plainFile.WriteString(resp.str + "\n")
} else {
    plainFile.WriteString(stripANSI(resp.str) + "\n")
}
```
it saves an unnecessary call to stripANSI when the user is using the -nc flag.

Mzack9999:
@ayanrajpoot10 Thanks for pointing out the optimization. I've summarized it into an helper function

Mzack9999 (approved):

PR #2297, opened 2025-10-22
fix: ensure Host header is set in unsafe mode for rawhttp requests

### LINKED ISSUE #2296: Host header overwritten in unsafe mode when using -target with raw requests
### httpx version:
latest (dev branch)

### Current Behavior:
When using `-rr` (raw request) together with `-target` in unsafe mode, httpx overwrites the Host header from the raw request file with the target host. This causes the original Host header specified in the request file to be ignored.

### Expected Behavior:
The Host header from the raw request file should be preserved, even when `-target` and `-unsafe` are used.

### Steps To Reproduce:
1. Create a raw request file `raw.txt`:
```http
GET / HTTP/1.1
Host: example.com
````

2. Run httpx with:

```bash
httpx -rr raw.txt -target test.com -unsafe -debug
```

3. Observe that the Host header is changed to `test.com` instead of staying as `example.com`.

ayanrajpoot10:
fixes #2296

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Host header synchronization in unsafe mode to ensure proper header handling.

### DIFF
--- a/common/httpx/httpx.go
+++ b/common/httpx/httpx.go
@@ -424,6 +424,9 @@ func (h *HTTPX) SetCustomHeaders(r *retryablehttp.Request, headers map[string]st
 		switch strings.ToLower(name) {
 		case "host":
 			r.Host = value
+			if h.Options.Unsafe {
+				r.Header.Set("Host", value)
+			}
 		case "cookie":
 			// cookies are set in the default branch, and reset during the follow redirect flow
 			fallthrough


Mzack9999 (changes requested):
@ayanrajpoot10 Thanks for your PR. This function is already performed automatically by the `rawhttp` library (ref: https://github.com/projectdiscovery/rawhttp/blob/132ebe7ef211155e77497e4432c5cb13dec1ba79/options.go#L31). Did you face any situation in which the action was not performed properly?

ayanrajpoot10:
> @ayanrajpoot10 Thanks for your PR. This function is already performed automatically by the `rawhttp` library (ref: https://github.com/projectdiscovery/rawhttp/blob/132ebe7ef211155e77497e4432c5cb13dec1ba79/options.go#L31). Did you face any situation in which the action was not performed properly?


image

image

In the first screenshot, raw.txt shows Host: de1.wstunnel.xyz, but when the request is made, it changes to arturo.ns.cloudflare.com, which is our target — that’s the issue. In the second screenshot, after making this change, the Host header remains as de1.wstunnel.xyz, which means the issue is resolved. This suggests that there may be something wrong with the rawhttp library.

ayanrajpoot10:
@Mzack9999 The problem is that `httpx.SetCustomHeaders` only set `req.Host` and didn’t add it to `req.Header`, so when `httpx.doUnsafeWithOptions `passed only `req.Header` to `rawhttp`, the Host header was missing, and `rawhttp.AutomaticHostHeader` overwrote it with the target URL’s host.

ayanrajpoot10:
`rawhttp` does not handle `req.Host` like Go’s standard HTTP package does; instead, it only uses `req.Header` to get the host value.

PR #2317, opened 2025-11-20
Adding panic guard + Tests

### LINKED ISSUE #2316: After updating httpx to the latest version, I’m consistently seeing a panic when httpx tries to classify certain HTML pages (likely deeply nested or malformed HTML).
Panic message:

panic: html: open stack of elements exceeds 512 nodes

Stack trace (from httpx):
github.com/projectdiscovery/httpx/common/pagetypeclassifier.htmlToText
github.com/projectdiscovery/httpx/common/pagetypeclassifier.(*PageTypeClassifier).Classify
github.com/projectdiscovery/httpx/runner.(*Runner).analyze
github.com/projectdiscovery/httpx/runner.(*Runner).process.func1
...

I am unsure of what triggered this but basically I was using httpx like this:
"cat hosts.txt| httpx -cl -sc -title -td -ct -server -vhost -o httpx-out.txt",

Now I have updated httpx on my 2 vps machine and both when doing this exhaustive lists of hosts, errored out like this so I do not know the actual url's that triggered this error. But in my localmachine where my httpx is still the older version, it worked completely fine.

I believe there is an issue in this new update, I am very bad at this hence I am really sorry if i have wasted your time in creating this issue if it is not legitimate, but this is what I noticed like after using httpx today morning and after updating it to the latest version.

Image

Mzack9999:
Closes #2316 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML sanitization and conversion to prevent crashes and parser overflow on deeply nested or malformed HTML.
  * Enhanced error handling so conversion failures return safe, empty results instead of causing failures.

* **Tests**
  * Added tests verifying recovery from deeply nested HTML and correct handling of normal and problematic HTML inputs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/common/pagetypeclassifier/pagetypeclassifier.go
+++ b/common/pagetypeclassifier/pagetypeclassifier.go
@@ -4,10 +4,11 @@ import (
 	_ "embed"
 	"sync"
 
-	"github.com/microcosm-cc/bluemonday"
-	htmltomarkdown "github.com/JohannesKaufmann/html-to-markdown/v2"
-	"github.com/projectdiscovery/utils/ml/naive_bayes"
 	"fmt"
+
+	htmltomarkdown "github.com/JohannesKaufmann/html-to-markdown/v2"
+	"github.com/microcosm-cc/bluemonday"
+	"github.com/projectdiscovery/utils/ml/naive_bayes"
 )
 
 //go:embed clf.gob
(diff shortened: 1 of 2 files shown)


dogancanbakir:
test fails

Issue #2345, opened 2025-12-27
`resume.cfg` may be written before the tool fully validates or flushes the current processing


ayuxsec:
### httpx version:

v1.7.4 (latest)

### Current Behavior:

After interrupting httpx with `CTRL+C`, resuming with `r resume.cfg` does resume execution, but some URLs that were not fully processed before interruption are skipped. The resumed run does not reprocess all remaining targets, resulting in missing output compared to normal run.

### Expected Behavior:

Running `httpx -r resume.cfg` should resume processing the remaining targets from the resume file and produce output accordingly.

---

### Steps To Reproduce:

1. Run `httpx` with input from stdin:

   ```bash
   cat subs.txt | httpx
   ```
2. Interrupt execution using `CTRL+C` after partial output.
3. Observe that `resume.cfg` is created.
4. Attempt to resume execution:

   ```bash
   cat subs.txt | httpx -r resume.cfg
   ```
5. Observer that some targets are not present compared to the output in normal run

### Anything else:

**Full terminal output:**

```console
ayux@pop-os:/tmp$ cat subs.txt | httpx

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

        projectdiscovery.io

[INF] Current httpx version v1.7.4 (latest)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
https://example.com
https://test.io
https://google.com
https://youtube.com
ayux@pop-os:/tmp$ cat subs.txt | httpx

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

        projectdiscovery.io

[INF] Current httpx version v1.7.4 (latest)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
https://example.com
https://test.io
^C[INF] CTRL+C pressed: Exiting
[INF] Creating resume file: resume.cfg
[ble: exit 1]
ayux@pop-os:/tmp$ cat subs.txt | httpx -r resume.cfg

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
[...]

PR #2388, opened 2026-02-02
fix: ensure resume.cfg is written after in-flight work completes

erdogan98:
/claim #2345

## Summary
This PR fixes the issue where `resume.cfg` was being written prematurely before in-flight HTTP requests had completed. This caused some URLs to be skipped when resuming with `-r resume.cfg`.

## Problem
When CTRL+C was pressed:
1. The signal handler immediately called `Close()` and `SaveResumeConfig()`
2. This happened while worker goroutines were still processing URLs
3. URLs that were in-flight were marked as "processed" in the resume file
4. On resume, these URLs were skipped even though they weren't actually processed

## Solution
1. **Added graceful shutdown** - Instead of calling `os.Exit(1)` immediately, the signal handler now calls `Interrupt()` which signals the runner to stop accepting new work
2. **Let in-flight work complete** - `RunEnumeration()` continues until all submitted work finishes (`wg.Wait()`)
3. **Write resume after completion** - `SaveResumeConfig()` is only called after `RunEnumeration()` returns, ensuring all submitted work has completed
4. **Check interrupt before incrementing** - The interrupt check happens before the resume index is incremented, so interrupted items are not marked as processed

## Changes
- `runner/runner.go`:
  - Added `shutdownChan`, `interrupted`, and mutex fields to Runner struct
  - Added `Interrupt()` and `IsInterrupted()` methods for graceful shutdown
  - Added interrupt check in `processItem` before incrementing resume index
- `cmd/httpx/httpx.go`:
  - Signal handler now calls `Interrupt()` instead of `os.Exit(1)`
  - Resume config is saved after `RunEnumeration()` completes
  - Exit code 1 is returned if interrupted

## Testing
Tested manually with the reproduction steps from the issue:
1. Run `cat subs.txt | ./httpx`
2. CTRL+C after partial output
3. Resume with `cat subs.txt | ./httpx -r resume.cfg`
4. All remaining targets are now properly processed

## Release Notes

* **Bug Fixes**
  * Graceful interrupt handling: the application now responds to Ctrl+C by signaling the runner to interrupt enumeration gracefully, with resume file creation deferred until enumeration finishes
  * Enhanced shutdown behavior: improved interrupt signaling and status code management to exit with error status only when the application is actually interrupted

### DIFF
--- a/cmd/httpx/httpx.go
+++ b/cmd/httpx/httpx.go
@@ -70,25 +70,37 @@ func main() {
 	}
 
 	// Setup graceful exits
+	// Instead of calling os.Exit immediately on SIGINT, we signal the runner
+	// to stop accepting new work and let in-flight work complete.
+	// This ensures resume.cfg is only written after all submitted work finishes.
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 	go func() {
 		for range c {
 			gologger.Info().Msgf("CTRL+C pressed: Exiting\n")
-			httpxRunner.Close()
-			if options.ShouldSaveResume() {
-				gologger.Info().Msgf("Creating resume file: %s\n", runner.DefaultResumeFile)
-				err := httpxRunner.SaveResumeConfig()
-				if err != nil {
-					gologger.Error().Msgf("Couldn't create resume file: %s\n", err)
-				}
-			}
-			os.Exit(1)
+			// Signal runner to stop accepting new work (but let in-flight work complete)
+			httpxRunner.Interrupt()
 		}
 	}()
 
 	httpxRunner.RunEnumeration()
+
+	// After RunEnumeration completes (including wg.Wait() for all in-flight work),
+	// save resume config if the run was interrupted
+	if httpxRunner.IsInterrupted() && options.ShouldSaveResume() {
+		gologger.Info().Msgf("Creating resume file: %s\n", runner.DefaultResumeFile)
+		err := httpxRunner.SaveResumeConfig()
+		if err != nil {
+			gologger.Error().Msgf("Couldn't create resume file: %s\n", err)
+		}
+	}
+
 	httpxRunner.Close()
+
+	// Exit with code 1 if interrupted
+	if httpxRunner.IsInterrupted() {
+		os.Exit(1)
+	}
 }
 
 // setupOptionalAssetUpload is used to setup optional asset upload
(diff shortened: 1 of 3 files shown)


erdogan98:
Addressed CodeRabbit feedback:

- **Removed dead fields**: Deleted `completedIndex`, `completedInput`, and `completedMu` from the Runner struct
- These fields were only being written to, never read - confirmed via grep that no code ever accessed them
- The `completedMu` mutex was solely protecting writes to these dead fields, so it was also removed
- The entire tracking block (lock/assignments/unlock) was removed as it served no purpose

The resume functionality continues to work correctly via `resumeCfg.currentIndex` which is properly maintained elsewhere in the codebase.

Build verified: `go build ./...` passes successfully.

erdogan98:
## Addressed CodeRabbit Feedback

This PR now addresses all the CodeRabbit feedback:

### Previously addressed (commit ab05874):
1. ✅ **Removed unused `completedIndex` and `completedInput` fields** - These were dead code that was never read
2. ✅ **Added nil check for `r.options.resumeCfg`** - Removed the code that accessed resumeCfg.currentIndex unconditionally (which would have caused a panic when running without resume)

### Newly addressed (commit 58b3b6b):
3. ✅ **Fixed goroutine leak in `streamInput()`** - The unbuffered channel could block indefinitely after `errInterrupted` caused the consumer to stop
   - Added `trySend()` helper that selects on both the output channel and `shutdownChan`
   - Now the goroutine exits cleanly when shutdown is signaled
   - Also added shutdown checks in the format.Parse callback

4. ⚠️ **HybridMap.Scan error propagation** - This is a limitation in the upstream `hmap` library. The `Scan` method swallows callback errors and always continues iteration. However, with the current implementation, `processItem` already checks for interrupt at the start of each iteration, so shutdown still works correctly - it just processes one more item before stopping. For truly large inputs, a separate PR to the hmap library would be needed to add cancellation support.

### Tests added:
- `TestInterruptFlag` - Verifies Interrupt() correctly sets the flag and closes shutdownChan
- `TestStreamInputShutdown` - Verifies streamInput goroutine exits on shutdown
- `TestTrySendBehavior` - Tests the trySend helper function behavior
- `TestResumeCfgNilSafety` - Verifies nil resumeCfg doesn't cause panics
- `TestInterruptDuringProcessItem` - Tests interrupt detection in processItem
- `TestGoroutineCleanupOnInterrupt` - Verifies no significant goroutine leaks

All tests pass locally.

dogancanbakir:

PR #2390, opened 2026-02-02
fix: prevent resume from skipping in-flight items

Simplereally:
## Summary

This PR fixes the resume functionality to prevent skipping items that were in-flight when Ctrl+C was pressed.

## Problem

When httpx is interrupted with Ctrl+C, the resume file saves `currentIndex` which includes items that were queued for processing but hadn't completed yet. On resume, these in-flight items would be skipped because their index was already counted, causing missing output.

**Root cause**: The `currentIndex` is incremented at the **start** of `processItem`, before `r.process()` spawns goroutines for the actual HTTP requests. When saving resume during interrupt, up to `Threads` items could be in-flight.

## Solution

When saving the resume config, subtract the thread count from `currentIndex` to get a "safe" index:

```go
safeIndex := r.options.resumeCfg.currentIndex - r.options.Threads
if safeIndex < 0 {
    safeIndex = 0
}
resumeCfg.Index = safeIndex
```

This is a conservative fix: it may re-process a few items that had already completed, but this is safe (httpx handles duplicates) while skipping items is not.

## Testing

- Added unit tests for the fix in `runner/resume_test.go`
- Manual testing: Verified resume now includes previously skipped items

## Checklist

- [x] Code compiles correctly
- [x] Tests pass
- [x] Documentation is up to date

/claim #2345

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced resume point calculation to safely handle concurrent requests in flight. The system now accounts for items still being processed, preventing potential data loss by adjusting where operations resume to avoid skipping unfinished requests.

* **Tests**
  * Added unit tests for resume configuration, including edge cases for concurrent thread handling and index boundary conditions.

### DIFF
--- a/runner/resume_test.go
+++ b/runner/resume_test.go
@@ -1,0 +1,51 @@
+package runner
+
+import (
+	"testing"
+)
+
+func TestSaveResumeConfigSubtractsThreads(t *testing.T) {
+	// Create a mock options with resume config
+	opts := &Options{
+		Threads: 10,
+		resumeCfg: &ResumeCfg{
+			currentIndex: 25,
+			current:      "https://example.com",
+		},
+	}
+
+	r := &Runner{
+		options: opts,
+	}
+
+	// Save resume config should subtract thread count to get safe index
+	// This ensures in-flight items are not skipped on resume
+	err := r.SaveResumeConfig()
+	if err != nil {
+		t.Fatalf("SaveResumeConfig failed: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


Simplereally:
Hi @Mzack9999 @ehsandeep 👋 This PR fixes the resume.cfg skip issue (#2345) with a thread-safe approach. Ready for review whenever you have a moment. Thanks! 🙏

Simplereally:
Closing in favor of #2391 which has proper mutex protection and more precise tracking. Will refine #2391 instead.

Issue #2014, opened 2024-11-27
Duplicate filter feature ideas (-fd, -filter-duplicates)


Xitro01:
Hello there,

I was surprised to see this feature implemented, as within my bug bounty hunting automation I used something similar like this:
```
subdomains_live.txt | ~/go/bin/httpx -sc -title -cl -wc -td -ip > subdomains_httpx_live.txt
subdomains_httpx_live.txt | awk -F"[" '!seen[$2, $3, $4, $5, $6]++ {print $1}' > subdomains_deduped.txt
```
Looking at how it is currently implemented, I have the feeling that a lot of potential good targets are going to be missed out on.
I believe there are many servers/applications out there that will respond exactly the same. Although there might be completely different content to be found.

So in my own automation I also included to look at the IP address. Because it is very likely if it is also from the same server IP address and the responses are exactly the same that it is actually just different subdomains pointing to the same webapplication.

With this I went from 5000 subdomains to just 500, which improves my chances to actually find something good. Because less is more in this case. Further analysis or automated scanning will take much less time (10% of the time) and you're not overwhelmed with all the duplicate results.

It might be a good idea to include this as well, or maybe there are others that have even better ideas.

GeorginaReeder:
Thanks so much for your feature request @Xitro01 ! We'll take a look into this :)

dogancanbakir:
Mentioning @Mzack9999 @ehsandeep for the discussion/brainstorming.

Xitro01:
Has been quite some time now. Let me help with some brainstorming.

I believe that when you filter out unique matches based on the following, you have a great chance to filter out true dupes.

All need to be checked to filter:

- IP
- CNAME
- Status Code (might not even be needed, due to Title/CL)
- Content Length (or Word Count)

Could be further refined with (but maybe takes too much time):

- Title
- Favicon
- Server / Tech Stack
- Location

buildingvibes:
/attempt #2014

I've implemented an enhancement to the duplicate filter feature that addresses this issue. The duplicate detection now considers both response content similarity AND IP address when determining duplicates.

**Key changes:**
- Responses are only filtered as duplicates if they have BOTH similar content AND the same IP address
- This ensures that identical responses from different servers (different IPs) are NOT filtered out
- Maintains backward compatibility with the existing `-fd` flag

**Pull Request:** https://github.com/projectdiscovery/httpx/pull/2400

[...]

PR #2400, opened 2026-02-09
Enhance duplicate filter to consider IP address

buildingvibes:
## Summary
This PR enhances the duplicate filtering feature (`-fd` flag) by incorporating the IP address into the duplicate detection logic, addressing issue #2014.

Previously, responses were considered duplicates based solely on content similarity using simhash. This approach had a limitation: identical responses from different IP addresses would be filtered out, potentially causing users to miss legitimate targets on different servers.

**Key Changes:**
- Modified `simHashes` cache to use composite key format: `"simhash:ip"` instead of just `simhash`
- Updated `duplicate()` function to compare both content similarity AND IP address
- Responses are now only considered duplicates if they have BOTH:
  1. Similar response content (simhash comparison with threshold ≤ 3)
  2. The same source IP address

## Problem Solved
As described in issue #2014, bug bounty hunters often encounter scenarios where different subdomains point to different servers (different IPs) but return identical responses. With the previous implementation, these would be incorrectly filtered as duplicates, causing users to miss potentially valuable targets.

Example scenario from the issue:
```bash
# Before: 5000 subdomains → 500 after deduplication
# But this could filter out legitimate different servers with same responses

# After: More precise deduplication
# Same response + same IP = duplicate (filtered)
# Same response + different IP = NOT duplicate (kept)
```

## Benefits
- **More precise duplicate detection** for bug bounty hunting workflows
- **Prevents filtering of legitimate targets** on different servers
- **Reduces false positives** in duplicate detection
- **Maintains backward compatibility** with existing `-fd` flag usage

## Test Coverage
Added comprehensive test `TestRunner_DuplicateFilterWithIP` that verifies:
1. Same content + same IP = duplicate (filtered)
2. Same content + different IP = NOT duplicate (kept)
3. Different content + same IP = NOT duplicate (kept)

All existing tests pass without modification.

## Example Usage
```bash
# Works exactly the same as before, but with improved precision
cat subdomains.txt | httpx -sc -title -cl -wc -td -ip -fd

# Now correctly keeps responses with same content but different IPs
```

/claim #2014

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced duplicate filtering to consider source IP addresses. Identical response content from different IP sources is no longer incorrectly flagged as duplicate, significantly improving detection accuracy across distributed systems and multiple endpoints.

* **Tests**
  * Added comprehensive test coverage for IP-aware duplicate filtering validation across multiple scenarios.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -94,7 +94,7 @@ type Runner struct {
 	browser            *Browser
 	pageTypeClassifier *pagetypeclassifier.PageTypeClassifier // Include this for general page classification
 	pHashClusters      []pHashCluster
-	simHashes          gcache.Cache[uint64, struct{}] // Include simHashes for efficient duplicate detection
+	simHashes          gcache.Cache[string, struct{}] // Include simHashes for efficient duplicate detection (key: "simhash:ip")
 	httpApiEndpoint    *Server
 	authProvider       authprovider.AuthProvider
 	interruptCh        chan struct{}
@@ -429,7 +429,7 @@ func New(options *Options) (*Runner, error) {
 		runner.HostErrorsCache = gc
 	}
 
-	runner.simHashes = gcache.New[uint64, struct{}](1000).ARC().Build()
+	runner.simHashes = gcache.New[string, struct{}](1000).ARC().Build()
 	pageTypeClassifier, err := pagetypeclassifier.New()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


buildingvibes:
Closing this PR to focus on one contribution at a time per project guidelines. I'll resubmit if bandwidth allows. Thank you\!

Issue #2240, opened 2025-08-13
-pr http11 flag is ignored on retryablehttp-go due to HTTP/2 fallback


khs-alt:
### Versions
- httpx: v1.7.1
- retryablehttp-go: v1.0.119

### Current Behavior: 
If you see httpx repo`common/httpx/httpx.go` 156 line
``` go
	if httpx.Options.Protocol == "http11" {
		// disable http2
		os.Setenv("GODEBUG", "http2client=0")
		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
	}
```
When using HTTP 1.1, change transport, and it will use retryablehttp
But if we see retryablehttp repo `do.go` 62 line
``` go
		// if err is equal to missing minor protocol version retry with http/2
		if err != nil && stringsutil.ContainsAny(err.Error(), "net/http: HTTP/1.x transport connection broken: malformed HTTP version \"HTTP/2\"", "net/http: HTTP/1.x transport connection broken: malformed HTTP response") {
			resp, err = c.HTTPClient2.Do(req.Request)
			checkOK, checkErr = c.CheckRetry(req.Context(), resp, err)
		}
```
This issue is caused by the retry behavior in retryablehttp-go, as discussed in
  projectdiscovery/retryablehttp-go#3.
If httpx set client for using only http1.1 but retryable use another client in this case so `-pr` does not work correctly.
### Expected Behavior:

Use transport change client when httpx `-pr` is on

### Suggestion

Add Options.DisableHTTPFallback (Example). If set, skip the HTTP/2 fallback (HTTPClient2) on malformed HTTP/2 errors and treat them according to the retry policy. 
This respects explicit HTTP/1.1-only configs and avoids unintended protocol switches and policy divergence      (proxy/dialer/redirect/TLS)

1234-ad:
## Fix Submitted! 🎉

I've analyzed the issue and submitted a comprehensive fix across two repositories:

### Root Cause
The `-pr http11` flag was being ignored because:
1. **httpx** correctly disables HTTP/2 in the main client
2. **retryablehttp-go** automatically falls back to HTTP/2 on certain errors, bypassing httpx's configuration

### Solution

**PR #521 to retryablehttp-go:** https://github.com/projectdiscovery/retryablehttp-go/pull/521
- Adds `DisableHTTP2` option to prevent automatic HTTP/2 fallback
- Fully backward compatible (disabled by default)

[...]

PR #2402, opened 2026-02-14
fix: prevent HTTP/2 fallback when -pr http11 is set

nenadilic84:
## Summary

When using `-pr http11`, httpx correctly disables HTTP/2 on the primary transport by clearing `TLSNextProto`. However, retryablehttp-go maintains a separate internal client (`HTTPClient2`) with HTTP/2 enabled, which it silently falls back to whenever the HTTP/1.1 request fails with a "malformed HTTP version" error from HTTP/2-only servers. This means the `-pr http11` flag has no effect against servers that only speak HTTP/2.

The fix overrides `HTTPClient2` on the retryablehttp client with an HTTP/1.1-only client after construction, so the fallback path also uses the same protocol-restricted transport.

## Root Cause

In `retryablehttp-go/do.go`, the retry loop catches HTTP/1.x transport errors caused by servers responding with HTTP/2 and transparently retries the request using `HTTPClient2`, which has `http2.ConfigureTransport` applied. This second client completely bypasses the transport configuration that httpx set up to enforce HTTP/1.1. Since `HTTPClient2` is a public field on the retryablehttp `Client` struct, the simplest fix is to replace it with a client that shares the same HTTP/1.1 transport.

## Changes

- `common/httpx/httpx.go`: After creating the retryablehttp client, when `Protocol == "http11"`, replace `HTTPClient2` with a new `http.Client` that uses the same HTTP/1.1-only transport (with `TLSNextProto` cleared). This ensures retryablehttp's internal fallback path does not upgrade to HTTP/2.

## Test plan

- All existing tests pass (`go test ./...`, `go vet ./...`)
- Verified by reading the retryablehttp-go source that the fallback in `do.go` line 63-66 uses `HTTPClient2.Do()`, and this change ensures that client is now constrained to HTTP/1.1 when requested

Fixes #2240

## Bug Fixes
* Fixed an issue where the HTTP/1.1 protocol flag was not consistently enforced across fallback clients, preventing silent fallback to HTTP/2 when explicitly disabled.

### DIFF
--- a/common/httpx/httpx.go
+++ b/common/httpx/httpx.go
@@ -183,6 +183,18 @@ func New(options *Options) (*HTTPX, error) {
 		CheckRedirect: redirectFunc,
 	}, retryablehttpOptions)
 
+	// When HTTP/1.1 is explicitly requested, override the retryablehttp fallback
+	// client (HTTPClient2) to also use the HTTP/1.1-only transport. Without this,
+	// retryablehttp falls back to HTTPClient2 (which has HTTP/2 enabled) when it
+	// encounters "malformed HTTP version" errors, silently ignoring the -pr http11 flag.
+	if httpx.Options.Protocol == "http11" {
+		httpx.client.HTTPClient2 = &http.Client{
+			Transport:     transport,
+			Timeout:       httpx.Options.Timeout,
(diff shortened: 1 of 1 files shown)


dogancanbakir:

PR #2409, opened 2026-02-17
fix: disable HTTP/2 fallback when -pr http11 is set

285729101:
## Summary
- Override retryablehttp-go's internal `HTTPClient2` fallback client with an HTTP/1.1-only transport when `-pr http11` is explicitly specified
- This prevents the library from silently falling back to HTTP/2 on `malformed HTTP version` errors, which was causing the `-pr http11` flag to be ignored
- The fix clones the existing HTTP/1.1 transport and sets `TLSNextProto` to an empty map to fully disable HTTP/2 negotiation on the fallback client

## Test plan
- [ ] Verify `go build ./...` compiles successfully
- [ ] Verify `go test ./common/httpx/...` passes
- [ ] Manual test: run `httpx -u <target> -pr http11` against an HTTP/2 server and confirm only HTTP/1.1 is used
- [ ] Manual test: confirm normal (non `-pr http11`) behavior is unaffected and HTTP/2 fallback still works

/claim #2240

🤖 Generated with [Claude Code](https://claude.com/claude-code)


285729101:
@Ice3man543 alternative approach for the HTTP/1.1 flag fix — disables HTTP/2 at the transport level.

dogancanbakir:

PR #2439, opened 2026-03-06
fix(runner): add mutex to testAndSet to prevent race condition

### LINKED ISSUE #2457: testAndSet race condition causes duplicate target processing
testAndSet in runner.go does a non-atomic check-then-set: it calls seen(k) and then setSeen(k) separately. Under concurrent access (e.g. TLS SubjectAN/SubjectCN discovery in processTargets), two goroutines can both pass seen() before either calls setSeen(), resulting in the same target being processed twice.

usernametooshort:
## Summary

Fixes a race condition in `testAndSet` where the read-check-then-write sequence is not atomic.

## Problem

`testAndSet` performs:
1. `seen(k)` - check if key exists
2. `setSeen(k)` - set the key

Without synchronization, two concurrent goroutines can both pass the `seen()` check before either calls `setSeen()`, causing duplicate processing of the same target.

## Impact

This race occurs in `processTargets()` where multiple goroutines call `testAndSet()` for discovered TLS SubjectAN/SubjectCN values (lines 1692, 1697, 1706, 1746, 1751).

When the race triggers:
- Same target gets processed multiple times
- Duplicate output entries
- Wasted HTTP requests and resources

## Fix

Add a `sync.Mutex` to the Runner struct to serialize `testAndSet` operations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the robustness of the deduplication system to ensure consistent behavior during concurrent operations.

* **Tests**
  * Added comprehensive unit tests for the deduplication method, including concurrent execution scenarios to validate correctness.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -80,6 +80,7 @@ import (
 
 // Runner is a client for running the enumeration process.
 type Runner struct {
+	seenMux sync.Mutex
 	options            *Options
 	hp                 *httpx.HTTPX
 	wappalyzer         *wappalyzer.Wappalyze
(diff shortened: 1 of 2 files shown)


Mzack9999 (changes requested):
The PR should point to the `dev` branch

PR #2447, opened 2026-03-09
fix(http11): prevent retryablehttp HTTP/2 fallback from bypassing -pr http11 (#2240)

marchantdev:
## Root Cause

When `-pr http11` is specified, httpx correctly disables HTTP/2 on the main transport:
- Sets `GODEBUG=http2client=0`
- Clears `transport.TLSNextProto`

However, `retryablehttp-go`'s `do.go` has an unconditional fallback at line 63–64:

```go
if err != nil && stringsutil.ContainsAny(err.Error(), "malformed HTTP version \"HTTP/2\"", ...) {
    resp, err = c.HTTPClient2.Do(req.Request)
}
```

`HTTPClient2` is a separate `*http.Client` configured with `golang.org/x/net/http2.ConfigureTransport`. Because it is set independently of the HTTP/1.1-only `HTTPClient`, this fallback silently bypasses the `-pr http11` restriction whenever a server responds with HTTP/2 framing.

## Fix

After constructing the `retryablehttp` client, when `Protocol == "http11"`, replace its `HTTPClient2` with the already-configured HTTP/1.1-only `HTTPClient`:

```go
if httpx.Options.Protocol == "http11" {
    httpx.client.HTTPClient2 = httpx.client.HTTPClient
}
```

This ensures the retryablehttp fallback also honours the HTTP/1.1-only constraint without requiring any upstream changes to `retryablehttp-go` and without affecting `client2`/`SupportHTTP2` for non-restricted clients.

## Tests

Added `TestHTTP11ProtocolEnforcement` with two sub-tests:

1. `http11 mode disables HTTPClient2 fallback` — asserts `HTTPClient2 == HTTPClient` when `Protocol == "http11"`
2. `default mode keeps distinct HTTPClient2` — asserts the two clients differ in default mode

```
=== RUN   TestHTTP11ProtocolEnforcement
=== RUN   TestHTTP11ProtocolEnforcement/http11_mode_disables_HTTPClient2_fallback
--- PASS (0.07s)
=== RUN   TestHTTP11ProtocolEnforcement/default_mode_keeps_distinct_HTTPClient2
--- PASS (0.05s)
--- PASS: TestHTTP11ProtocolEnforcement (0.12s)
PASS
```

## Checklist

- [x] Targets `dev` branch
- [x] Minimal change — only applies when `Protocol == "http11"`
- [x] No upstream retryablehttp-go changes required
- [x] Unit tests pass
- [x] `go build ./...` passes
- [x] Fixes #2240

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures HTTP/1.1 mode is consistently enforced during HTTP client setup, preventing unintended HTTP/2 fallbacks.

* **Tests**
  * Added test coverage to validate HTTP/1.1 enforcement and to confirm default protocol behavior remains unchanged.

### DIFF
--- a/common/httpx/httpx.go
+++ b/common/httpx/httpx.go
@@ -183,6 +183,15 @@ func New(options *Options) (*HTTPX, error) {
 		CheckRedirect: redirectFunc,
 	}, retryablehttpOptions)
 
+	// When HTTP/1.1-only mode is enforced, prevent retryablehttp-go's HTTP/2 fallback
+	// from bypassing the protocol restriction. retryablehttp-go falls back to HTTPClient2
+	// (an HTTP/2-capable client) when it encounters a "malformed HTTP version" error from
+	// a server that speaks HTTP/2. Replacing HTTPClient2 with the HTTP/1.1-only HTTPClient
+	// ensures the -pr http11 flag is honoured end-to-end. See: #2240
+	if httpx.Options.Protocol == "http11" {
+		httpx.client.HTTPClient2 = httpx.client.HTTPClient
+	}
+
 	transport2 := &http2.Transport{
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: true,
--- a/common/httpx/httpx_test.go
+++ b/common/httpx/httpx_test.go
@@ -2,6 +2,7 @@ package httpx
 
 import (
 	"net/http"
+	"os"
 	"testing"
 
 	"github.com/projectdiscovery/retryablehttp-go"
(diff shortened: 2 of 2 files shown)


marchantdev:
Addressed CodeRabbit review:

- **Pointer-identity assertions**: Replaced `require.Equal/NotEqual` with `require.Same/NotSame` on `ht.client.HTTPClient` vs `ht.client.HTTPClient2` to correctly assert pointer identity rather than value equality.
- **GODEBUG isolation**: Added `t.Setenv("GODEBUG", os.Getenv("GODEBUG"))` in each subtest so the `New` call's GODEBUG mutation in the first subtest doesn't pollute the second's global state. `t.Setenv` automatically restores the original value on test cleanup.

dwisiswant0 (approved):

Mzack9999:
This issue has been resolved by merging #2424. While the core fix here is correct, it uses the string literal `"http11"` instead of the `HTTP11` constant from `proto.go`, and the tests add unnecessary `GODEBUG` environment save/restore logic that isn't needed for pointer identity assertions. Closing as superseded. Thank you for the contribution.

PR #2448, opened 2026-03-09
fix: ensure -pr http11 disables HTTP/2 fallback

innerpeace609:
This PR fixes issue #2240 where the -pr http11 flag was being ignored due to the underlying transport falling back to HTTP/2. I have explicitly disabled ForceAttemptHTTP2 and cleared TLSNextProto in the transport configuration when the http11 protocol is selected.

/claim #2240

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced HTTP protocol configuration to prevent unintended HTTP/2 connections in HTTP/1.1 mode.

### DIFF
--- a/common/httpx/httpx.go
+++ b/common/httpx/httpx.go
@@ -157,6 +157,7 @@ func New(options *Options) (*HTTPX, error) {
 		// disable http2
 		_ = os.Setenv("GODEBUG", "http2client=0")
 		transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}
+		transport.ForceAttemptHTTP2 = false
 	}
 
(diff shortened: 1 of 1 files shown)


Mzack9999:
Thank you for looking into this! Unfortunately, this change doesn't address the root cause. The issue is that `retryablehttp-go` internally falls back to a separate `HTTPClient2` (with HTTP/2 enabled) on certain transport errors — setting `ForceAttemptHTTP2 = false` on the primary transport is redundant since `TLSNextProto` already disables HTTP/2 there. The fix (overriding `HTTPClient2` to reuse the HTTP/1.1 client) has been merged in #2424. Closing this PR.

PR #2456, opened 2026-03-19
fix: respect -pr http11 during retryablehttp fallback

drmabus:
/claim #2240
Closes #2240

Summary

When `-pr http11` is specified, httpx should strictly use HTTP/1.1.

However, retryablehttp may fall back to HTTPClient2 (which has HTTP/2 enabled)
on certain transport errors, bypassing the user's explicit protocol selection.

## Fix

If HTTP/1.1 is forced, ensure HTTPClient2 reuses the same HTTP/1.1-only client,
preventing unintended HTTP/2 fallback.

## Impact

- Respects explicit protocol selection
- Prevents silent HTTP/2 usage
- Minimal and safe change

## Testing

Verified locally:

- go build ./...
- ./httpx -u https://example.com -pr http11 -debug
- ./httpx -u https://nghttp2.org -pr http11 -debug
- ./httpx -u https://cloudflare.com -pr http11 -debug

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced HTTP/1.1 for configured client connections so primary and retry clients share the same transport, preventing unexpected protocol fallback during retries.

* **Tests**
  * Added tests confirming that HTTP/1.1 configuration disables protocol fallback and that default settings preserve the existing retry behavior.

### DIFF
--- a/common/httpx/httpx.go
+++ b/common/httpx/httpx.go
@@ -183,6 +183,13 @@ func New(options *Options) (*HTTPX, error) {
 		CheckRedirect: redirectFunc,
 	}, retryablehttpOptions)
 
+	// When http11 is forced, override HTTPClient2 to also use the HTTP/1.1-only
+	// transport. Without this, retryablehttp silently falls back to HTTPClient2
+	// (which has HTTP/2 enabled) on certain transport errors, bypassing -pr http11.
+	if httpx.Options.Protocol == "http11" {
+		httpx.client.HTTPClient2 = httpx.client.HTTPClient
+	}
+
 	transport2 := &http2.Transport{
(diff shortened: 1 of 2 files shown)


drmabus:
Thanks for reviewing! Let me know if you'd like any changes or additional test cases.

drmabus:
Updated to use the HTTP11 constant instead of string literals for consistency. Thanks for the suggestion.

Mzack9999:
Thank you for the contribution! This issue has been resolved by merging #2424, which implements the same approach (`HTTPClient2 = HTTPClient`). Closing as the fix is now in `dev`. Appreciate the clean implementation and tests.

PR #2491, opened 2026-05-11
feat: add -sh / -security-headers probe for HSTS and X-Content-Type-Options

technical-faults-dev:
**Added new security headers probe:**

- New flag: `-sh` or `-security-headers`
- Detects `Strict-Transport-Security` (HSTS)
- Detects `X-Content-Type-Options: nosniff`

Useful for bug bounty hunters and security assessments.

Part of ProjectDiscovery OSS Bounty Program.


dwisiswant0:
Since this is a brand new feature, it'd be awesome if you could open up an issue first so we have a bit of a discussion and discourse with the rest of the community and make sure we're all on the same page before moving forward.

Also, it looks like there's already a pretty simple workaround for this using the `-filter-condition`/`-fdc` or `-match-condition`/`-mdc` flags, so let's chat about whether we actually need to bake this in.

Issue #2495, opened 2026-05-15
-fpt error does not work in v1.9.0


ipsecc:
### v1.8.1
```
└─# ./httpx -u jsdk.baidu.com -fep

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.8.1 (outdated)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
                                                                           
```
<img width="684" height="301" alt="Image" src="https://github.com/user-attachments/assets/cb67a65d-cab3-42c3-acb4-51b0756b6b62" />

### v1.9.0
As you can see, the error page is not filtered.
```
└─# ./httpx -u jsdk.baidu.com -fpt error

    __    __  __       _  __
   / /_  / /_/ /_____ | |/ /
  / __ \/ __/ __/ __ \|   /
 / / / / /_/ /_/ /_/ /   |
/_/ /_/\__/\__/ .___/_/|_|
             /_/

                projectdiscovery.io

[INF] Current httpx version v1.9.0 (latest)
[WRN] UI Dashboard is disabled, Use -dashboard option to enable
https://jsdk.baidu.com
```
<img width="719" height="297" alt="Image" src="https://github.com/user-attachments/assets/d0535970-db2a-4045-ad27-55c989487294" />

singhvishalkr:
/attempt #2495

I'll investigate and fix the -fpt error filter regression introduced in v1.9.0. The filter was working correctly in v1.8.1 but no longer filters error pages in the latest version.

Plan:
1. Reproduce the issue locally
2. Identify the regression in the page classification/filter logic
3. Fix the filter to properly match error page types
4. Add tests to prevent future regressions

Will submit a PR shortly.

dogancanbakir:
Fixed it in https://github.com/HappyHackingSpace/dit/pull/48. We just need to bump version in httpx, after mege-n-release in dit.

PR #2496, opened 2026-05-16
fix(runner): return error when -fpt filter fails to initialize

singhvishalkr:
## Summary
- When users specify \-fpt\ (filter-page-type) flag, the dit classifier must be successfully initialized
- Previously, if the dit model was missing, httpx would only log a warning and silently fail to apply the filter
- This caused unexpected behavior where error pages would not be filtered as expected

## Changes
- Return a descriptive error when the classifier cannot be initialized for \-fpt\ usage
- Include guidance for users to run \dit data download\ to install the required model

## Proof

**Before (silent failure):**
\\\
httpx -u jsdk.baidu.com -fpt error
https://jsdk.baidu.com  # Error page NOT filtered
\\\

**After (clear error):**
\\\
httpx -u jsdk.baidu.com -fpt error
FATAL: could not initialize page classifier for -fpt filter: dit: model.json not found (run 'dit data download' to install the model)
\\\

## Checklist
- [x] PR created against the correct branch (main)
- [x] Code follows project conventions
- [x] Fix is minimal and focused

/claim #2495

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -434,6 +434,9 @@ func New(options *Options) (*Runner, error) {
 	if options.JSONOutput || options.CSVOutput || len(options.OutputFilterPageType) > 0 {
 		ditClassifier, err := dit.New()
 		if err != nil {
+			if len(options.OutputFilterPageType) > 0 {
+				return nil, fmt.Errorf("could not initialize page classifier for -fpt filter: %w (run 'dit data download' to install the model)", err)
+			}
 			gologger.Warning().Msgf("Could not initialize page classifier: %s", err)
 		}
 		runner.ditClassifier = ditClassifier


Mzack9999:
Thanks for the attempt, but this doesn't actually fix #2495.

The reported bug is that `-fpt error` no longer filters error pages out of the box (it worked in v1.8.1 via `-fep` with zero setup). This PR only converts the silent `dit.New()` failure into a fatal error and points users at `dit data download` — a CLI that isn't documented anywhere in this repo. After this change, `httpx -u ... -fpt error` still doesn't filter anything on a clean install; it just refuses to run, which is a UX regression for existing users with working `-fpt` invocations in scripts/CI.

The real fix needs to live in the `dit` bootstrap path — embed the model in the package, ship it with httpx releases, or auto-fetch it into `~/.dit/` on first use (similar to how nuclei pulls templates). Closing this in favor of a proper fix for #2495.

PR #2553, opened 2026-08-10
Pin workflow actions and fix the codeql-action advisory exposure

Totara-thib:
## Proposed changes

Three commits, nothing about how the workflows behave changes except one security bump.

- **Move codeql-action off the advisory-affected v2 line.** The v2 releases from 2.26.11 on sit in the range of [GHSA-vqf5-2xx6-9wfm](https://github.com/advisories/GHSA-vqf5-2xx6-9wfm), CodeQL debug artifacts could contain the GitHub PAT, and no fixed v2 release exists, the fix shipped in 3.28.3. The v2 line is also deprecated. The three references move to the current release, pinned by SHA.
- **Pin every workflow action to a full commit SHA.** The release path runs goreleaser with a token that can create releases, and the docker workflow logs into Docker Hub with the registry credentials. A version tag is a movable pointer, whoever controls an action repository can point it at different code after review, which is exactly what the tj-actions/changed-files incident (CVE-2025-30066) did to leak CI secrets. A SHA cannot be retargeted. Each pin keeps the version as a trailing comment for cross-checking against the action's releases page, versions stay exactly where they were, and dependabot keeps bumping SHA pins the same way it bumps tags today. This also settles the codeql refs resolving ambiguously as both a tag and a branch. The `projectdiscovery/actions` pins are included for consistency, happy to drop those hunks if you prefer trusting your own org namespace.
- **Declare token permissions.** Only the release workflow writes through the GitHub token (goreleaser creating the release), it keeps `contents: write`. The test and docker workflows drop to read only, the Docker Hub push authenticates with its own registry credentials.

One heads-up: if the org restricts allowed actions with tag patterns like `owner/action@v2`, those patterns stop matching SHA refs and workflows fail at startup, the fix is `owner/action@*` in that setting. Nothing to do if no such restriction is configured.

### Proof

Every SHA resolves to the exact tag named in its comment, verifiable on each action's releases page. All workflow files parse and the changes are ref-format only, except the codeql bump described above.

## Checklist

- [x] Pull request is created against the [dev](https://github.com/projectdiscovery/httpx/tree/dev) branch
- [x] All checks passed (lint, unit/integration/regression tests etc.) with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works *(not applicable, workflow-only change)*
- [ ] I have added necessary documentation (if appropriate) *(not applicable)*

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.

### DIFF
--- a/.github/workflows/auto-merge.yaml
+++ b/.github/workflows/auto-merge.yaml
@@ -18,11 +18,11 @@ jobs:
     runs-on: ubuntu-latest
     if: github.actor == 'dependabot[bot]'
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           token: ${{ secrets.DEPENDABOT_PAT }}
 
-      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@45fc124d949b19b6b8bf6645b6c9d55f4f9ac61a # v2.6.6
         with:
           github-token: ${{ secrets.DEPENDABOT_PAT }}
           target: all
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -7,6 +7,9 @@ on:
       - '**.mod'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   lint:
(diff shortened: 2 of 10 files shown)


dogancanbakir:
we dont need pinned hashes if there are versioning in those actions, which is the case. closing this, please consider using versioning for them.

PR #2570, opened 2026-08-29
test(runner): add custom port fallback test cases to TestSwitchPortForFallback

gcoinstash-cmd:
### Summary of Changes
- Adds test cases in `TestSwitchPortForFallback` for non-default custom ports (e.g. port `3000` on HTTP and port `5000` on HTTPS).
- Verifies that non-default custom port numbers remain unchanged during protocol fallback attempts.

### Test Validation
- `go test -v ./runner -run TestSwitchPortForFallback`: **6/6 test cases passed 100% green in 1.213s**.


Mzack9999:
Thanks for the interest. These target `main` while we land on `dev`, and GitHub is showing the whole `dev` delta rather than a small test change. A few of the later commits are also empty. Please close the extras and, if you still want a focused test PR, open one against `dev` with only the new tests.

PR #2571, opened 2026-08-30
test(httputilz): add test case for query parameter preservation in ParseRequest

gcoinstash-cmd:
### Summary of Changes
- Adds a unit test case `TestParseRequestPreservesQueryParameters` to verify that query strings in raw HTTP request lines (e.g. `/api/v1/users?page=2&limit=50`) and authentication headers are accurately extracted and preserved without distortion.

### Test Validation
- `go test -v ./common/httputilz -run TestParseRequestPreservesQueryParameters`: **passed 100% green in 0.523s**.


Mzack9999:
Thanks for the interest. These target `main` while we land on `dev`, and GitHub is showing the whole `dev` delta rather than a small test change. A few of the later commits are also empty. Please close the extras and, if you still want a focused test PR, open one against `dev` with only the new tests.

PR #2577, opened 2026-09-02
fix: stop reporting TLS-only ports as plain http

### LINKED ISSUE #2580: Reduce latency from unsuccessful HTTPS upgrade after plaintext HTTP 400
## Summary

The HTTPS upgrade logic added in [PR #2577](https://github.com/projectdiscovery/httpx/pull/2577) retries an automatic plaintext HTTP probe that receives an HTTP 400 response. If the target is cleartext-only or otherwise does not support HTTPS, the HTTPS retry can consume the full probe timeout before httpx restores the original plaintext HTTP result.

## Required changes

Evaluate and implement a bounded strategy that reduces or avoids the latency from an HTTPS retry that cannot succeed. Preserve the scheme-detection behavior for TLS-only services that initially return HTTP 400 to a plaintext probe.

## Rationale

The existing retry is correct for TLS-only services, but an unsuccessful retry adds avoidable latency for affected targets.

## Affected areas

- `runner/runner.go`: HTTPS upgrade and fallback logic in `analyze`.
- `runner/runner_test.go`: Regression coverage for cleartext HTTP 400 responses and unsuccessful HTTPS upgrade attempts.

## Acceptance criteria

- A plaintext HTTP 400 result remains available when the HTTPS upgrade attempt fails.
- TLS-only services that return HTTP 400 to a plaintext probe continue to be reported as HTTPS.
- Cleartext HTTP 400 targets do not incur an unbounded or full normal probe-timeout delay from the HTTPS upgrade attempt.
- Tests cover both successful TLS upgrade and unsuccessful HTTPS upgrade behavior.

Requested by @dwisiswant0.

Backlinks:
- Pull request: https://github.com/projectdiscovery/httpx/pull/2577
- Request comment: https://github.com/projectdiscovery/httpx/pull/2577#issuecomment-5532983050

knakul853:
## Change
A port above 1024 is probed as plain HTTP first, and the scheme retry only fires on a transport error. A TLS-only listener answers a plaintext request with a perfectly valid HTTP 400, so `err == nil`, the retry never happens, and the service is reported as plain http with a 400, no title and no technologies.

- Trigger on the shape of the exchange — a plaintext probe answered 400 — and let the TLS handshake settle it. No parsing of the server's error text, so every phrasing is covered.
- Uses its own flag rather than the shared `retried`, so when TLS does not work the existing scheme fallback still recovers the original http result.
- Skipped in unsafe mode, which bypasses that fallback and would drop the result instead of recovering it.

## Evidence
- `TestTLSOnlyPortIsProbedOverHTTPS`: real runner against an in-process TLS-only listener. `dev` reports `scheme=http`/400; this reports `scheme=https`/200.
- `TestPlainHTTPPortStaysHTTP`: the other direction — a genuine cleartext 400 must stay http. This is what makes the text-free trigger safe.
- Compared against unmodified `dev` across 7 server shapes (TLS-only, explicit http, explicit https, plain 200, plain 400, unsafe, probe): exactly one changed, the bug.
- Integration suite run locally; the two `rfc path` cases cover the unsafe behaviour.

## Still open
- A cleartext service answering 400 now costs 3 connections instead of 1 — still reported as http, and it fails fast. Services returning anything other than 400 are untouched at 1 connection.
- A port answering 400 over both schemes is now reported as https rather than http.
- Restricting the trigger to `*443` ports would avoid the extra connections, but misses TLS on arbitrary high ports, so it was not taken.

Claude-Session: https://claude.ai/code/session_015AKXsNzjd3HjGMq4B8JCNH
Claude-Local-Session: claude --resume 0fe5b680-5ba5-43e6-b354-43d4388aaf9c

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling during automatic HTTP-to-HTTPS upgrades.
  * Preserves successful HTTP results and URLs when an HTTPS attempt fails, without repeating the request.
  * Correctly distinguishes genuine HTTP 400 responses from services that require HTTPS.
  * Maintains accurate results for one-time HTTP services when subsequent connections are refused.
  * Restores plaintext results when TLS connections close without returning an HTTP response.

Open in Codex: codex://threads/01a06453-c2da-7800-9217-0af3676d92d5
Resume from CLI: codex resume 01a06453-c2da-7800-9217-0af3676d92d5


knakul853:
## Reproduction

Self-contained, stdlib only. It serves HTTPS on `:19443` and answers plaintext the way a real TLS listener does — a valid HTTP 400 saying TLS is required, byte-identical to the response that triggered this (62-byte CRLF body).

Save as `tlsonly.go`, run `go run tlsonly.go`, then in another shell:

```console
$ httpx -u 127.0.0.1:19443 -silent -nc -duc
http://127.0.0.1:19443        # released httpx v1.9.0 - wrong, the port is TLS-only

$ curl -sk https://127.0.0.1:19443/ | head -1
<html><head><title>Only Over TLS</title></head><body>secret</body></html>

$ curl -s http://127.0.0.1:19443/ | tail -1
This combination of host and port requires TLS.
```

The port is provably TLS - curl gets the page over https - yet httpx records `scheme=http` with a 400, no title and no technologies. Build this branch and the same command reports `https://127.0.0.1:19443`.

Note: `origin/dev` does not currently build (`tlsx v1.3.2` is not a real tag). To compile: `go mod edit -replace github.com/projectdiscovery/tlsx@v1.3.2=github.com/projectdiscovery/tlsx@v1.3.0`

<details>
<summary><code>tlsonly.go</code></summary>

```go
// tlsonly serves HTTPS on :19443 and answers any plaintext request the way a
// real TLS listener does: a valid HTTP 400 saying TLS is required.
// Run: go run tlsonly.go
package main

import (
	"bufio"
	"crypto/rand"
	"crypto/rsa"
	"crypto/tls"
	"crypto/x509"
	"crypto/x509/pkix"
	"io"
	"log"
	"math/big"
	"net"
	"net/http"
	"time"
)

const rejection = "HTTP/1.1 400 Bad Request\r\n" +
	"Connection: close\r\n" +
	"Content-Type: text/plain;charset=utf-8\r\n" +
	"Content-Length: 62\r\n\r\n" +
	"Bad Request\r\nThis combination of host and port requires TLS.\r\n"

func main() {
	key, err := rsa.GenerateKey(rand.Reader, 2048)
	if err != nil {
		log.Fatal(err)
	}
	tmpl := x509.Certificate{
		SerialNumber: big.NewInt(1),
		Subject:      pkix.Name{CommonName: "tls-only.test"},
		DNSNames:     []string{"localhost", "tls-only.test"},
		NotBefore:    time.Now().Add(-time.Hour),
		NotAfter:     time.Now().Add(24 * time.Hour),
		KeyUsage:     x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
	}
	der, err := x509.CreateCertificate(rand.Reader, &tmpl, &tmpl, &key.PublicKey, key)
	if err != nil {
		log.Fatal(err)
	}
	cfg := &tls.Config{Certificates: []tls.Certificate{{Certificate: [][]byte{der}, PrivateKey: key}}}

	mux := http.NewServeMux()
	mux.HandleFunc("/", func(w http.ResponseWriter, _ *http.Request) {
		w.Header().Set("Content-Type", "text/html")
		_, _ = io.WriteString(w, "<html><head><title>Only Over TLS</title></head><body>secret</body></html>")
	})
	srv := &http.Server{Handler: mux, ReadHeaderTimeout: 5 * time.Second}

	ln, err := net.Listen("tcp", "127.0.0.1:19443")
	if err != nil {
		log.Fatal(err)
	}
	log.Println("TLS-only service on 127.0.0.1:19443 — plaintext gets HTTP 400")

	for {
		conn, err := ln.Accept()
		if err != nil {
			return
		}
		go func(conn net.Conn) {
			_ = conn.SetReadDeadline(time.Now().Add(5 * time.Second))
			br := bufio.NewReader(conn)
			first, err := br.Peek(1)
			if err != nil {
				_ = conn.Close()
				return
			}
			// 0x16 is a TLS handshake record; anything else is plaintext.
			if first[0] != 0x16 {
				drain(br)
				_, _ = io.WriteString(conn, rejection)
				_ = conn.Close()
				return
			}
			_ = srv.Serve(&oneShot{c: tls.Server(&peeked{Conn: conn, r: br}, cfg)})
		}(conn)
	}
}

func drain(r io.Reader) {
	sc := bufio.NewScanner(r)
	for sc.Scan() {
		if sc.Text() == "" {
			return
		}
	}
}

type peeked struct {
	net.Conn
	r *bufio.Reader
}

func (p *peeked) Read(b []byte) (int, error) { return p.r.Read(b) }

type oneShot struct {
	c    net.Conn
	used bool
}

func (o *oneShot) Accept() (net.Conn, error) {
	if o.used {
		return nil, io.EOF
	}
	o.used = true
	return o.c, nil
}
func (o *oneShot) Close() error   { return nil }
func (o *oneShot) Addr() net.Addr { return o.c.LocalAddr() }
```

</details>

dwisiswant0:
@coderabbitai create an issue to track this.

dwisiswant0 (approved):

Issue #1494, opened 2023-12-20
Get Chinese garbled title


b1win0y:
I use the httpx program under the terminal to get the title as normal Chinese, but when I get the title through code, it is garbled.
my httpx version is 1.3.7 and I use the following code to get the title and get garbled characters

```
package main

import (
	"fmt"
	"log"
	"github.com/projectdiscovery/goflags"
	"github.com/projectdiscovery/gologger"
	"github.com/projectdiscovery/gologger/levels"
	"github.com/projectdiscovery/httpx/runner"
)

func main() {
	gologger.DefaultLogger.SetMaxLevel(levels.LevelVerbose)
	options := runner.Options{
		Methods:         "GET",
		InputTargetHost: goflags.StringSlice{"http://103.45.249.230:7878/"},
		//InputFile: "./targetDomains.txt", // path to file containing the target domains list
		OnResult: func(r runner.Result) {
			fmt.Println("Title:", r.Title)
			fmt.Println("Raw:", r.Raw)
			// handle error
			if r.Err != nil {
				fmt.Printf("[Err] %s: %s\n", r.Input, r.Err)
				return
			}
			fmt.Printf("%s %s %d\n", r.Input, r.Host, r.StatusCode)
		},
	}
	if err := options.ValidateOptions(); err != nil {
		log.Fatal(err)
	}
	httpxRunner, err := runner.New(&options)
	if err != nil {
		log.Fatal(err)
	}
	defer httpxRunner.Close()
	httpxRunner.RunEnumeration()
}
```
<img width="1090" alt="image" src="https://github.com/projectdiscovery/httpx/assets/49974970/cf8c0acf-33df-4e76-94a7-abaade577d63">
and  the Chinese content of the body is also garbled，Please tell me how to solve this problem？

I think it may be related to the gb2312 encoding

acidvegas:
This is simple a lack of fonts on your system. If you want to actually see the unicode, ensure you have fonts that support high level unicode. Some of the fonts i installed are:
`font-unifont-bdf freefont-ttf noto-fonts-ttf noto-fonts-ttf-extra noto-fonts-cjk noto-fonts-emoji`

Also, MOST terminals are using utf8 by default, but make sure your locales are setup to us en_US UTF-8 or w.e in /etc/locale or whereever its managed on your system.

Mzack9999:
Closing as the issue seems unrelated to the tool

Issue #652, opened 2022-06-05
GBK encode Issue 


burpheart:
```console
echo 'https://rce.moe/testgbk.html' | httpx -json -irr | jq .
```

```json
{
  "timestamp": "2022-06-09T03:29:09.567107+05:30",
  "request": "GET /testgbk.html HTTP/1.1\r\nHost: rce.moe\r\nUser-Agent: MOT-V9mm/00.62 UP.Browser/6.2.3.4.c.1.123 (GUI) MMP/2.0\r\nAccept-Charset: utf-8\r\nAccept-Encoding: gzip\r\n\r\n",
  "response-header": "HTTP/1.1 200 OK\r\nConnection: close\r\nContent-Length: 175\r\nAccept-Ranges: bytes\r\nAccess-Control-Allow-Origin: *\r\nAge: 21732\r\nCache-Control: public, max-age=0, must-revalidate\r\nContent-Disposition: inline; filename=\"testgbk.html\"\r\nContent-Type: text/html; charset=utf-8\r\nDate: Wed, 08 Jun 2022 21:59:03 GMT\r\nEtag: \"2c7a0e73141df4a888bcf5db9d2c2271\"\r\nServer: Vercel\r\nStrict-Transport-Security: max-age=63072000\r\nX-Vercel-Cache: HIT\r\nX-Vercel-Id: bom1:bom1::tqv77-1654725543465-1fb2ce41502c\r\n\r\n",
  "scheme": "https",
  "port": "443",
  "path": "/testgbk.html",
  "a": [
    "76.76.21.22"
  ],
  "url": "https://rce.moe:443/testgbk.html",
  "input": "https://rce.moe/testgbk.html",
  "title": "测试测试测试测试",
  "webserver": "Vercel",
  "response-body": "<!DOCTYPE html>\n<html lang=\"zh-CN\">\n<head>\n<title>���Բ��Բ��Բ���</title>\n<meta charset=\"GBK\">\n\n<body>\n<h1>\n���Բ��Բ��Բ��Բ��Բ��Բ��Բ��Բ��Բ��Բ��Բ��Բ���\n</h1>\n</body>",
  "content-type": "text/html",
  "method": "GET",
  "host": "76.76.21.22",
  "content-length": 175,
  "status-code": 200,
  "response-time": "3.124937709s",
  "failed": false,
  "hashes": {
    "body-md5": "2c7a0e73141df4a888bcf5db9d2c2271",
    "body-mmh3": "361964648",
[...]

Issue #441, opened 2021-11-25
The software only supports charset="UTF-8"


pandak1d:
The software only supports charset="UTF-8". If the title of charset="euc-kr" will appear garbled, is there a golang package that can automatically convert the encoding format according to charset, just like the library in python
```
mychar = chardet.detect(response.content)
response.encoding = mychar['encoding']
```
![image](https://user-images.githubusercontent.com/69457324/143391484-d01865b1-ca9d-475f-88d2-a821137e492b.png)
![image](https://user-images.githubusercontent.com/69457324/143392359-29d0a007-6ac9-45a0-8151-788b5d14c6c3.png)

ehsandeep:
```console
echo https://www.100yearshop.co.kr/shop/main/index.php | httpx -title -silent -timeout 20
https://www.100yearshop.co.kr/shop/main/index.php []
```

pandak1d:
Like this is garbled！！！
![image](https://user-images.githubusercontent.com/69457324/144558368-c8d253eb-5ff5-469a-8dca-efb17b13ec18.png)

pandak1d:
Why was it removed

Issue #37, opened 2020-07-10
bug with output http title


Malayke:
Hi, I find a bug, httpx can not correct output none ASCII string, like Chinese

```
▶ echo https://tnw.qq.com  | httpx -title -silent 
https://tnw.qq.com [The Next Web����վ_��Ѷ�Ƽ�_��Ѷ��]
```
I try to fix it by myself, but I'm not good at go

the correct output is:
```
echo https://tnw.qq.com  | httpx -title -silent | iconv -f gbk
https://tnw.qq.com [The Next Web中文站_腾讯科技_腾讯网]
```

bimkos:
I don't know what will happen to other requests. Do you think this will be a universal solution? Can I add a switch for encodings?

bauthard:
@bimkos, adding a flag for this will increase the complexity, as a "user" we don't know when we going to encounter for the cases like this, so we can't be sure when to use a flag or not, so surely we need to come up with something which works in existing flow.

bauthard:
```bash
> echo tnw.qq.com  | ./httpx -title -silent -follow-redirects
https://tnw.qq.com [The Next Web中文站_腾讯科技_腾讯网]
```

@Malayke this is fixed now.

PR #2591, opened 2026-09-06
fix: handle quoted charset and non-canonical content-type in title extraction

krishna28238-arch:
httpx silently drops the title, or extracts it as mojibake, when the `Content-Type` response header deviates from the canonical lowercase form:

- `Content-Type: text/html; charset="gbk"` — the quoted charset value (legal per RFC 9110) is not matched by the `charset=gbk` substring check in `DecodeData`, so GBK/EUC-KR bodies are never transcoded and the title comes out as raw bytes
- `Content-Type: TEXT/HTML` or `text/html ; charset=utf-8` — media types are case-insensitive and optional whitespace is allowed before parameters, but `CanHaveTitleTag` compares the value verbatim, so the title extraction is skipped entirely

```
$ httpx -u http://127.0.0.1:8000 -title
# content-type: text/html; charset="gbk"        -> []  (raw GBK bytes)
# content-type: TEXT/HTML                       -> (no title at all)
```

This strips quotes from the joined content-type before the charset matching, and normalizes the mime type (case + surrounding whitespace) in `CanHaveTitleTag`. Responses with canonical headers are unaffected. Regression tests cover both the quoted-charset and case/whitespace variants end-to-end (`TestCanHaveTitleTag`, `TestExtractTitleDecodesCharset`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of quoted character-set values in HTTP content types, ensuring titles encoded with GBK and similar formats display correctly.
  - MIME type detection now handles capitalization differences and surrounding whitespace, improving recognition of HTML content.

- **Tests**
  - Added coverage for MIME type normalization and decoding titles from quoted and unquoted character-set declarations.

### DIFF
--- a/common/httpx/encodings.go
+++ b/common/httpx/encodings.go
@@ -58,6 +58,8 @@ func DecodeData(data []byte, headers http.Header) ([]byte, error) {
 	// Non UTF-8
 	if contentTypes, ok := headers["Content-Type"]; ok {
 		contentType := strings.ToLower(strings.Join(contentTypes, ";"))
+		// the charset parameter value can be a quoted string (charset="gbk")
+		contentType = strings.ReplaceAll(contentType, `"`, "")
 
 		switch {
 		case stringsutil.ContainsAny(contentType, "charset=gb2312", "charset=gbk"):
--- a/common/httpx/title.go
+++ b/common/httpx/title.go
@@ -51,7 +51,7 @@ func ExtractTitle(r *Response) (title string) {
 }
 
 func CanHaveTitleTag(mimeType string) bool {  
-    return slices.Contains(supportedTitleMimeTypes, mimeType)  
+    return slices.Contains(supportedTitleMimeTypes, strings.ToLower(strings.TrimSpace(mimeType)))  
 }  
 
 func getTitleWithDom(r *Response) (*html.Node, error) {
--- a/common/httpx/title_test.go
+++ b/common/httpx/title_test.go
@@ -1,0 +1,72 @@
+package httpx
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/projectdiscovery/retryablehttp-go"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCanHaveTitleTag(t *testing.T) {
+	tests := []struct {
+		mimeType string
+		expected bool
+	}{
+		{"text/html", true},
+		{"TEXT/HTML", true},
+		{"Text/Html", true},
+		{"text/html ", true},
+		{" application/xhtml+xml", true},
+		{"text/plain", false},
+		{"application/json", false},
+		{"", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.mimeType, func(t *testing.T) {
+			require.Equal(t, tt.expected, CanHaveTitleTag(tt.mimeType))
+		})
+	}
+}
+
+func TestExtractTitleDecodesCharset(t *testing.T) {
+	options := DefaultOptions
+	options.CdnCheck = "false"
+	options.Timeout = 2 * time.Second
+	options.RetryMax = 0
+
+	ht, err := New(&options)
+	require.Nil(t, err)
+
+	// <title>中文</title> with the title text encoded in GBK
+	gbk := []byte{0xd6, 0xd0, 0xce, 0xc4}
+	body := append([]byte("<html><head><title>"), gbk...)
+	body = append(body, []byte("</title></head></html>")...)
+
+	tests := []struct {
+		name        string
+		contentType string
+	}{
+		{"quoted charset", `text/html; charset="gbk"`},
+		{"unquoted charset", "text/html; charset=gbk"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+				w.Header().Set("Content-Type", tt.contentType)
+				_, _ = w.Write(body)
+			}))
+			defer srv.Close()
+
+			req, err := retryablehttp.NewRequest(http.MethodGet, srv.URL, nil)
+			require.Nil(t, err)
+			resp, err := ht.Do(req, UnsafeOptions{})
+			require.Nil(t, err)
+			require.Equal(t, "中文", ExtractTitle(resp))
+		})
+	}
+}


ehsandeep (commented):
@krishna28238-arch thanks for the PR, do you have test target to see before / after behavior?

krishna28238-arch:
Yes — I've included end-to-end regression tests that 
demonstrate the before/after behavior:

**Test 1: TestCanHaveTitleTag** (title_test.go)
Shows that non-canonical MIME types like "TEXT/HTML" and 
" text/html ; charset=utf-8" are now correctly recognized 
as HTML (previously skipped entirely, resulting in no 
title extraction).

**Test 2: TestExtractTitleDecodesCharset** (title_test.go)  
Shows that a GBK-encoded body with a quoted charset 
declaration `Content-Type: text/html; charset="gbk"` now 
transcodes correctly and extracts the title. Previously 
the quoted value failed the charset= substring match, so 
the body was never transcoded and the title came out as 
raw bytes.

You can verify the before/after behavior by:

1. Check out the main branch and run:
   go test ./common/httpx/ -run TestCanHaveTitleTag -v
   go test ./common/httpx/ -run TestExtractTitleDecodesCharset -v
   
   (Tests will fail or show incorrect behavior)

2. Check out this PR branch and run the same commands:
   (Tests will pass, showing correct behavior)

Or run the full test suite to confirm no regressions:
   go test ./common/httpx/ -v

ehsandeep:
@krishna28238-arch I'm looking for real hosts to better understand the existing tool behavior, then test code to better understand the change. Can you share some?

krishna28238-arch:
Thanks @ehsandeep, sure - here's what I found. The short version: the mainstream CJK web has moved to canonical UTF-8 headers, the legacy encodings are still out there, and the two shapes this PR fixes are straightforward to reproduce locally even though I couldn't find a live public host serving them today.

**Real hosts I checked**

I ran `curl -s -D -` over ~55 CN/KR/JP hosts (portals, provincial gov sites, university BBS front-ends, older reference sites). Every mainstream one is canonical now:

```
 $ curl -s -D - -o /dev/null https://www.baidu.com | grep -i ^content-type
Content-Type: text/html; charset=utf-8

 $ curl -s -D - -o /dev/null http://www.gov.cn | grep -i ^content-type
Content-Type: text/html
```

(baidu.com, qq.com, 163.com, sohu.com, sina.com.cn, gov.cn, people.com.cn, cctv.com, china.com.cn, naver.com, daum.net, yahoo.co.jp, etc)

One host that is still GB2312 today, in case you want to watch the existing behavior end to end:

```
 $ curl -s -D - -o /dev/null http://www.cnr.cn | grep -i ^content-type
Content-Type: text/html                        <- no charset in the header

 $ curl -s http://www.cnr.cn | grep -aoiE '<meta[^>]*charset[^>]*>' | head -1
<meta http-equiv="Content-Type" content="text/html; charset=gb2312">

 $ curl -s http://www.cnr.cn | file -bi -
text/html; charset=iso-8859-1                  <- body is GBK bytes
```

dev gets this one right - the `<meta>` fallback in `DecodeData` catches `charset=gb2312"`:

```
 $ git checkout dev && go build -o /tmp/httpx-dev ./cmd/httpx
 $ echo http://www.cnr.cn | /tmp/httpx-dev -title -silent -no-color
https://www.cnr.cn [央广网·中央广播电视总台]
```

So the meta-declared path works today. What doesn't work is when the same value comes back in the HTTP header in non-canonical form:

1. quoted charset, `text/html; charset="gbk"` - `DecodeData` substring-matches `charset=gbk`, misses the quoted form, so the GBK body never gets transcoded and the title comes out as raw GBK bytes
2. MIME portion not exactly lowercase/trimmed - `TEXT/HTML`, or `text/html ` coming from `text/html ; charset=...` - `CanHaveTitleTag` does an exact `slices.Contains`, so title extraction is just skipped

This same class of host has come up in the tracker a few times already: #37 (tnw.qq.com, GBK, 2020), #441 (euc-kr titles, 2021), #652 (GBK page, 2022), #1494 (103.45.249.230:7878, 2023). All four are migrated or offline now, so for the code-test part I set up local servers that emit the exact shapes:

```python
# save as weird_ct.py
from http.server import HTTPServer, BaseHTTPRequestHandler
import threading, signal

class Handler(BaseHTTPRequestHandler):
    body = b''
    ctype = ''
    def do_GET(self):
        self.send_response(200)
        self.send_header('Content-Type', self.ctype)
        self.send_header('Content-Length', str(len(self.body)))
        self.end_headers()
        self.wfile.write(self.body)
    def log_message(self, fmt, *args):
        pass

class QuotedGBK(Handler):
    ctype = 'text/html; charset="gbk"'
    body = '<html><head><title>这是中文测试标题</title></head><body>Content</body></html>'.encode('gbk')

class UppercaseMIME(Handler):
    ctype = 'TEXT/HTML'
    body = b'<html><head><title>Uppercase MIME Type Test</title></head><body>Content</body></html>'

class Whitespace(Handler):
    ctype = 'text/html ; charset=utf-8'   # note the space before the semicolon
    body = b'<html><head><title>Whitespace Test</title></head><body>Content</body></html>'

for port, h in [(8001, QuotedGBK), (8002, UppercaseMIME), (8003, Whitespace)]:
    threading.Thread(target=HTTPServer(('127.0.0.1', port), h).serve_forever, daemon=True).start()
signal.pause()
```

```
 $ curl -s -D - -o /dev/null http://127.0.0.1:8001/ | grep -i ^content-type
Content-Type: text/html; charset="gbk"
 $ curl -s -D - -o /dev/null http://127.0.0.1:8002/ | grep -i ^content-type
Content-Type: TEXT/HTML
 $ curl -s -D - -o /dev/null http://127.0.0.1:8003/ | grep -i ^content-type
Content-Type: text/html ; charset=utf-8
```

dev branch:

```
 $ echo http://127.0.0.1:8001/ | /tmp/httpx-dev -title -silent -no-color
http://127.0.0.1:8001/ [ĲԱ]  <- mojibake, raw GBK passed through, never transcoded

 $ echo http://127.0.0.1:8002/ | /tmp/httpx-dev -title -silent -no-color
http://127.0.0.1:8002/       <- no title at all, TEXT/HTML not recognized

 $ echo http://127.0.0.1:8003/ | /tmp/httpx-dev -title -silent -no-color
http://127.0.0.1:8003/       <- no title, "text/html " (trailing space) not recognized
```

and to confirm the first one is really just undecoded GBK (the same bytes the server sent):

```
 $ echo http://127.0.0.1:8001/ | /tmp/httpx-dev -title -silent -no-color | iconv -f gbk -t utf-8
http://127.0.0.1:8001/ [这是中文测试标题]
```

with this PR:

```
 $ git checkout fix-title-content-type-edge-cases && go build -o /tmp/httpx-fixed ./cmd/httpx
 $ echo http://127.0.0.1:8001/ | /tmp/httpx-fixed -title -silent -no-color
http://127.0.0.1:8001/ [这是中文测试标题]

 $ echo http://127.0.0.1:8002/ | /tmp/httpx-fixed -title -silent -no-color
http://127.0.0.1:8002/ [Uppercase MIME Type Test]

 $ echo http://127.0.0.1:8003/ | /tmp/httpx-fixed -title -silent -no-color
http://127.0.0.1:8003/ [Whitespace Test]
```

**Regression check**

I ran both builds over the 32 reachable hosts from the scan with `-title -silent` and the titles came out identical. The only differences were a redirect race on cctv.com/naver.com (each run caught a different side of the 302, in opposite directions) and one transient timeout on xhby.net. `go test ./common/httpx/` passes on the branch, including the new `TestCanHaveTitleTag` / `TestExtractTitleDecodesCharset`.

